PR #1, opened 2017-12-29
Add initial plugin scaffolding

econchick:

Sets up how the plugin paths need to be when installed with [gordon](https://github.com/spotify/gordon). 

This also adds a dev dependency to auto-reload when docs are edited.

All the names and the repo organization in general is meant to be a placeholder for what we actually decide on in terms of naming convention, interface structure, etc.

N/A

This will fail because it makes an assumption that interfaces are defined in gordon core when they're not yet.

**Release note**:

```release-note
NONE
```

cc @spotify/alf


docs/providers.rst
```diff
@@ -0,0 +1,111 @@
...
+   compute
+   dns
+   pubsub
+
+
+Example Gordon Configuration with GCP Plugin
+--------------------------------------------
+
+.. literalinclude:: ../gordon.toml.example
+   :language: ini
```
econchick: The `literalinclude` sphinx directive includes the file itself and highlights it according to `language` defined (from pygments).

docs/providers.rst
```diff
@@ -0,0 +1,111 @@
...
+
+
+Plugin Configuration
+--------------------
+
+.. _gcp:
```
econchick: These are custom references. Instead of saying `:ref:[gcp]` (and maybe needing to escape some backticks) I can just say `:ref:gcp` when I want to link here (which is done below).

docs/providers.rst
```diff
@@ -0,0 +1,111 @@
...
+.. _gcpgce:
+
+``[gcp.compute]``
+~~~~~~~~~~~~~~~~~
+
+.. note::
+
+    Configuration defined for the ``compute`` provider will overwrite values of the same keys defined under :ref:`gcp <gcp>`, then inherit the rest.
+
+    The ``compute`` provider does **not** have access to configuration for :ref:`gcp.dns <gcpdns>` and :ref:`gcp.pubsub <gcppubsub>`.
```
smstone: just a suggestion but might want to just add a note at the top saying this, rather than saying it in every description?

econchick: ah yeah good call.

econchick:
@smstone thanks for the review - I just moved up the `note` to the beginning of the plugin config and removed the individual ones. I also changed it from a `note` to `attention` so it's more obvious (red background instead of gray):

2018-01-02 at 9 30 am

I'm going to merge this since you gave the +1, but let me know if you think anything else needs to be changed.

PR #2, opened 2018-01-03
[RFC] Add generic & audit log schemas

econchick:

Add message schemas for validation when subscribing to Google PubSub. This ports over the GCP-related aspects of @descent3me [original PR](https://github.com/spotify/gordon/pull/4). 

Validation will happen in gordon core (PR to come).

I'm adding documentation, but I've opened this up for comments for now.

N/A

Tests still aren't going to pass yet because of some initial assumptions made in previous PRs.

I've tested validation of the included examples with the `jsonschema` package locally:

```python
>>> import json
>>> import jsonschema
>>>
>>> with open('gordon_gcp/schemas/event.schema.json', 'r') as f:
...   event_schema = json.load(f)
>>> with open('gordon_gcp/schemas/examples/event.A.json', 'r') as f:
...   a_event_exp = json.load(f)
>>> with open('gordon_gcp/schemas/examples/event.NS.json', 'r') as f:
...   ns_event_exp = json.load(f)
>>> with open('gordon_gcp/schemas/examples/event.CNAME.json', 'r') as f:
...   cname_event_exp = json.load(f)
>>>
>>> # would throw errors if not valid
>>> jsonschema.validate(a_event_exp, event_schema)
>>> jsonschema.validate(ns_event_exp, event_schema)
>>> jsonschema.validate(cname_event_exp, event_schema)
>>>
>>> with open('gordon_gcp/schemas/audit-log.schema.json', 'r') as f:
...   audit_schema = json.load(f)
>>> with open('gordon_gcp/schemas/examples/audit-log.first-operation.json', 'r') as f:
...   first_audit_exp = json.load(f)
>>> with open('gordon_gcp/schemas/examples/audit-log.last-operation.json', 'r') as f:
...   last_audit_exp = json.load(f)
>>> with open('gordon_gcp/schemas/examples/audit-log.noop-operation.json', 'r') as f:
...   noop_audit_exp = json.load(f)
>>>
>>> jsonschema.validate(first_audit_exp, event_schema)
>>> jsonschema.validate(last_audit_exp, event_schema)
>>> jsonschema.validate(noop_audit_exp, event_schema)
```

**Release note**:

```release-note
NONE
```

cc @spotify/alf

### DIFF
--- a/gordon_gcp/schemas/audit-log.schema.json
+++ b/gordon_gcp/schemas/audit-log.schema.json
@@ -1,0 +1,57 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "type": "object",
+    "id": "https://github.com/spotify/gordon-gcp/blob/master/gordon_gcp/schemas/audit-log.schema.json",
+    "title": "Google Audit Log Messages",
+    "description": "Schema for parsing Google Audit Log Messages",
+    "required": ["protoPayload", "logName"],
+    "properties": {
+        "protoPayload": {
+            "type": "object",
+            "properties": {
+                "methodName": {
+                    "type": "string",
+                    "description": "Action performed on a GCP",
+                    "enum": [
+                        "v1.compute.instances.insert",
+                        "v1.compute.instances.delete"
+                    ]
+                },
+                "resourceName": {
+                    "type": "string",
+                    "description": "",
+                    "examples": [
+                        "projects/a-project-id/zones/a-zone-name/instances/an-instance-name"
+                    ],
+                    "pattern": "^projects/[0-9]+/zones/[a-z\\-0-9]+/instances/[a-z\\-0-9]+"
+                }
+            },
+            "required": ["methodName", "resourceName"]
+        },
+        "logName": {
+            "type": "string",
+            "description": "Name of audit log",
+            "examples": [
+                "projects/a-project-id/logs/cloudaudit.googleapis.com%2Factivity"
+            ],
+            "pattern": "^projects/[a-z][a-z\\-0-9]{5,29}/logs/cloudaudit.googleapis.com%2Factivity"
+        },
+        "receiveTimestamp": {
+            "type": "string",
+            "description": "Time the log entry was received by Stackdriver Logging",
+            "examples": [
+                "2018-01-01T23:13:45.123456789Z"
+            ],
+            "pattern": "^(\\d+)-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01])[Tt]([01][0-9]|2[0-3]):[0-5][0-9]:([0-5][0-9]|60)(\\.\\d+)([Zz]|([\\+|\\-]([01][0-9]|2[0-3]):[0-5][0-9]))"
+
+        },
+        "timestamp": {
+            "type": "string",
+            "description": "Time the operation status was reported",
+            "examples": [
+                "2018-01-01T23:13:45.123456789Z"
+            ],
+            "pattern": "^(\\d+)-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01])[Tt]([01][0-9]|2[0-3]):[0-5][0-9]:([0-5][0-9]|60)(\\.\\d+)([Zz]|([\\+|\\-]([01][0-9]|2[0-3]):[0-5][0-9]))"
+        }
+    }
+}
--- a/gordon_gcp/schemas/event.schema.json
+++ b/gordon_gcp/schemas/event.schema.json
@@ -1,3 +1,64 @@
 {
-    "foo": "bar"
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "type": "object",
+    "id": "https://github.com/spotify/gordon-gcp/blob/master/gordon_gcp/schemas/event.schema.json",
+    "title": "Generic Event Messages",
+    "description": "Schema for parsing generic event messages",
+    "required": ["action", "timestamp", "resourceRecords"],
+    "properties": {
+        "action": {
+            "description": "Add or delete resource record sets (rrdatas) listed",
+            "type": "string",
+            "enum": ["additions", "deletions"]
+        },
+        "timestamp": {
+            "type": "string",
+            "description": "Time the message was emitted",
+            "examples": [
+                "2018-01-01T23:13:45.123456789Z"
+            ],
+            "pattern": "^(\\d+)-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01])[Tt]([01][0-9]|2[0-3]):[0-5][0-9]:([0-5][0-9]|60)(\\.\\d+)([Zz]|([\\+|\\-]([01][0-9]|2[0-3]):[0-5][0-9]))"
+        },
+        "resourceRecords": {
+            "description": "DNS resource records upon which to act",
+            "type": "object",
+            "required": ["name", "rrdatas", "type"],
+            "properties": {
+                "name": {
+                    "description": "Record name",
+                    "type": "string",
+                    "examples": [
+                        "www.example.com.",
+                        "some-fqdn.subdomain.example.com."
+                    ]
+                },
+                "rrdatas": {
+                    "description": "Resource record(s) that `name` point to",
+                    "type": "array",
+                    "minItems": 1,
+                    "items": {
+                        "type": "string"
+                    },
+                    "uniqueItems": true,
+                    "examples": [
+                        "198.1.1.50",
+                        "example.com.",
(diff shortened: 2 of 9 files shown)


gordon_gcp/schemas/audit-log.schema.json
```diff
@@ -0,0 +1,56 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
```
econchick: There's drafts 6 and 7 out, but `jsonschema` only supports 3 and 4 at the moment.

gordon_gcp/schemas/audit-log.schema.json
```diff
@@ -0,0 +1,56 @@
...
+        },
+        "logName": {
+            "type": "string",
+            "description": "Name of audit log",
+            "examples": [
+                "projects/a-project-id/logs/cloudaudit.googleapis.com%2Factivity"
+            ],
+            "pattern": "^projects/[a-z][a-z\\-0-9]{5,29}/logs/cloudaudit.googleapis.com%2Factivity"
+        },
+        "receiveTimestamp": {
```
econchick: I figured these two could be useful for idempotence, but I was unsure so I left them as optional.

gordon_gcp/schemas/audit-log.schema.json
```diff
@@ -0,0 +1,56 @@
...
+            },
+            "required": ["methodName", "resourceName"]
+        },
+        "logName": {
+            "type": "string",
+            "description": "Name of audit log",
+            "examples": [
+                "projects/a-project-id/logs/cloudaudit.googleapis.com%2Factivity"
+            ],
+            "pattern": "^projects/[a-z][a-z\\-0-9]{5,29}/logs/cloudaudit.googleapis.com%2Factivity"
```
econchick: [Docs on project id format](https://cloud.google.com/resource-manager/reference/rest/v1/projects#Project)

gordon_gcp/schemas/audit-log.schema.json
```diff
@@ -0,0 +1,56 @@
...
+        },
+        "receiveTimestamp": {
+            "type": "string",
+            "description": "Time the log entry was received by Stackdriver Logging",
+            "examples": [
+                "2018-01-01T23:13:45.123456789Z"
+            ],
+            "pattern": "^([0-9]+)-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01])[Tt]([01][0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9]|60)(\\.[0-9]+)?(([Zz])|([\\+|\\-]([01][0-9]|2[0-3]):[0-5][0-9]))"
```
econchick: I suck at regexes - is there a better one for RFC 3339 format? Per [log entry operation docs](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#LogEntryOperation).

gordon_gcp/schemas/event.schema.json
```diff
@@ -1,3 +1,49 @@
...
+    "properties": {
+        "name": {
+            "description": "Record name",
+            "type": "string",
+            "examples": [
+                "www.example.com.",
+                "some-fqdn.subdomain.example.com."
+            ]
+        },
+        "rrdatas": {
```
econchick: I'm thinking technically this isn't needed if it's a `deletions` action. I can't quite [discern](https://cloud.google.com/dns/api/v1/changes#resource) if it's required for deleting. If it is, we can adjust the schema, or first call the DNS API to list the record sets. Thoughts?

gordon_gcp/schemas/event.schema.json
```diff
@@ -1,3 +1,49 @@
...
+                "198.1.1.50",
+                "example.com.",
+                "some text value"
+            ]
+        },
+        "ttl": {
+            "description": "TTL in seconds; default 300 seconds",
+            "type": "number",
+            "maximum": 86400,
+            "minimum": 300,
```
econchick: This minimum is based off of our current defaults.

gordon_gcp/schemas/event.schema.json
```diff
@@ -1,3 +1,49 @@
...
+        },
+        "type": {
+            "description": "Type of resource record(s) as supported by Google Cloud DNS",
+            "type": "string",
+            "enum": ["A", "AAAA", "CNAME", "MX", "NS", "PTR", "SOA", "SRV", "TXT"]
+        },
+        "action": {
+            "description": "Add or delete resource record sets (rrdatas) listed",
+            "type": "string",
+            "enum": ["additions", "deletions"]
```
econchick: Mimicking the [changes](https://cloud.google.com/dns/api/v1/changes#resource) endpoint.

gordon_gcp/schemas/examples/audit-log.first-operation.json
```diff
@@ -0,0 +1,54 @@
+{
+  "insertId": "123456789101112",
+  "logName": "projects/some-project/logs/cloudaudit.googleapis.com%2Factivity",
+  "operation": {
```
econchick: This `"operation"` key could be interesting; it could potentially help us avoid unnecessary polling Compute Engine if we know that if there's ever an `"operation"` key, that there will be at least 2 messages, one with a `"first"` key set to true, and another with `"last"` key set to true. If that's the case, we can shorten the polling of GCE, as well as only act if the operation was completed. The [docs](https://cloud.google.com/logging/docs/audit/api/#large_or_long-running_audit_log_entries) are a little unclear.

gordon_gcp/schemas/event.schema.json
```diff
@@ -1,3 +1,49 @@
 {
-    "foo": "bar"
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "type": "object",
+    "id": "https://github.com/spotify/gordon-gcp/blob/master/gordon_gcp/schemas/event.schema.json",
+    "title": "Generic Event Messages",
```
econchick: The idea for this second schema is for messages that are not automatically produced by Google's Audit Log of operations (i.e. manual records, a reconciliation process, etc).

gordon_gcp/schemas/event.schema.json
econchick: Ok I'll add `rrdatas` as required. 

At least with TTL we can default it to 300 seconds if not given when we're actually processing the event.

gordon_gcp/schemas/examples/audit-log.first-operation.json
descent3me: I'm not sure but I *think* the one with no operations gets issued at the same time as the "last".  I was only looking at the sample data (which of course I didn't think to save anywhere to I can find it now), so I'm not sure if it's guaranteed that way, but I liked that that data type was simpler, and just by excluding the ones with operations we could ensure getting only one message per instance.

gordon_gcp/schemas/event.schema.json
```diff
@@ -1,3 +1,49 @@
...
+    "description": "Schema for parsing generic event messages",
+    "required": ["name", "rrdatas", "action", "type"],
+    "properties": {
```
descent3me: I also wonder if we should have a timestamp in this message as well.  I don't like the complexity it could add to Gordon, but given that we have only additions and deletions, an update to a record would need to be one deletion and one addition, in that order, and since order isn't necessarily guaranteed over pubsub, I can't think of a way besides timestamp to ensure it.

econchick: yeah good call. It does look like messages pulled from pubsub will have a [publish_time](https://googlecloudplatform.github.io/google-cloud-python/latest/pubsub/subscriber/api/message.html) attribute (when using the python client library). But I also think it may be good for the publisher itself to add its own timestamp.

gordon_gcp/schemas/audit-log.schema.json
econchick: The [docs](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#top_of_page) (I linked to a wrong anchor id earlier - sorry!) to refer to the [Timestamp](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Timestamp) type that conforms to [RFC3339](https://www.ietf.org/rfc/rfc3339.txt), which does allow both `T`/`Z` and `t`/`z`.

The `\d+` does indeed work \o/ and the other suggested cleanups seem to work just fine too :D @descent3me I'm coming to you for all my regex needs from now on 😉

econchick:
@descent3me @matt0bi I've updated the PR with the requested/suggested changes (I hope I got everything). Mind taking another look?

matt0bi:
LGTM :+1:

PR #6, opened 2018-03-27
Add pub/sub message parser 

econchick:

When working on the event consumer implementation, I figured it'd be preferable to keep logic that is specific to message schemas (i.e. parsing) outside of the consumer itself. It makes it a bit more extensible; for example, if someone wanted to use/add a different schema, they'd just need to address the validation/parsing of it, and not have to worry about the actual message consumption.

The relevant message workflow I have in mind is:
1. Consume a message from Pub/Sub
2. Validate the message's data
    - `ack` the message + log if not valid against any schema
    - return the schema name if it is valid
3. Parse message according to schema
4. Create an event message (via the implementation of `IEventMessage`) based off of the data
5. Add it to the success channel for further processing
(not to mention the acking/extending msg deadline stuff)

The idea is that the event message (a provider of `IEventMessage`) may have missing information in the `data` attribute (to be filled in in the enricher), parsed via this parser. If the event consumer receives a general `event` message, then no further enriching is needing.

N/A

For test fixtures, I used the examples already included in the repo. Hopefully it makes sense.

**Release note**:

```release-note
NONE
```

@spotify/alf


gordon_gcp/schema/parse.py
```diff
@@ -0,0 +1,89 @@
...
+        payload = message['protoPayload']
+        data = {
+            'action': self.ACTION_MAPPER[payload['methodName']],
+            'resourceName': payload['resourceName'],
+            'resourceRecords': [],
+        }
+        return data
+
+    def _parse_event_msg(self, message):
+        # TODO (lynn): I think this needs to be a list
```
descent3me: I don't understand this TODO - was there something accidentally behind in the method or why is there a `records` var?

gordon_gcp/schema/parse.py
```diff
@@ -0,0 +1,89 @@
...
+        Args:
+            message (dict): message data to parse.
+            schema (str): valid message schema.
+        Returns:
+            (dict): parsed message
+        """
+        func = {
+            'audit-log': self._parse_audit_log_msg,
+            'event': self._parse_event_msg,
+        }[schema]
```
descent3me: Could this map be a object var like the action mapper (I'm unclear on the role of `self` here)?

gordon_gcp/schema/parse.py
```diff
@@ -0,0 +1,89 @@
...
+class MessageParser:
+    """Parse a message provided a given GCP schema."""
+    ACTION_MAPPER = {
+        'v1.compute.instances.insert': 'additions',
+        'v1.compute.instances.delete': 'deletions',
+    }
+
+    def _parse_audit_log_msg(self, message):
+        # TODO (lynn): what about first vs last messages?
+        payload = message['protoPayload']
```
descent3me: Don't we lose some info here by only taking the proto payload, like timestamp? Or would that be handled outside of this parser?

gordon_gcp/schema/parse.py
```diff
@@ -0,0 +1,89 @@
...
+        return {
+            'action': message['action'],
+            'resourceName': None,
+            'resourceRecords': [message['resourceRecords']]
+        }
+
+    def parse(self, message, schema):
+        """Parse message according to schema.
+
+        `message` should already be validated against the given schema.
```
descent3me: Is there any reason to do that separately?  I could see that as part of the parsing process.

gordon_gcp/schema/parse.py
```diff
@@ -0,0 +1,89 @@
...
+        return data
+
+    def _parse_event_msg(self, message):
+        # TODO (lynn): I think this needs to be a list
+        records = []
+        record = message['resourceRecords']
+        records.append(record)
+        return {
+            'action': message['action'],
+            'resourceName': None,
```
descent3me: Is this field really needed?  Could it be solved by adding a default field to the schema to make this method basically a noop?

gordon_gcp/schema/parse.py
econchick: My todo was unclear, but it's meant to potentially update the schema definition for `event.schema.json`. It looks like what google returns for DNS records is a list of resource records.

gordon_gcp/schema/parse.py
econchick: It needs self to have access to the methods that this class defines.

gordon_gcp/schema/parse.py
econchick: I wasn't sure exactly how timestamp would be used quite yet as I'm writing this consumer, but I can add it in with the anticipation of eventually using it. Doesn't matter to me.

gordon_gcp/schema/parse.py
econchick: It could, but the way I'm approaching it is to first validate it using the schema.validate.MessageValidator (and fail fast if it doesn't validate against any schema), then to parse it. I think I prefer the separation of validation then parsing. But open to hear arguments otherwise.

gordon_gcp/schema/parse.py
econchick: I originally did this for consistency between both schema-type parsing and to hint at what data would be needed to "enrich", but as I'm writing the consumer and trying to think about the enricher, I don't think there's any benefit to the consistency. I'll remove.

econchick:
@descent3me I'm open to a different name. If it helps, here's a slimmed down code snippet of how it'll be used in the consumer:

```py
async def _handle_pubsub_msg(self, pubsub_msg):
    data = json.loads(pubsub_msg.data)
    schema = self._get_and_validate_pubsub_msg_schema(data)
    # handle if it's not valid against any schema

    event_msg_data = self._parser.parse(data, schema)
    # create an event message obj (this is in the next PR, it implements `IEventMessage`)
    event_msg = self._create_gevent_msg(pubsub_msg, event_msg_data, schema)

    # if the message has resource records, assume it has all info
    # it needs to be published, and therefore is already enriched
    phase = 'enrich' if event_msg_data['resourceRecords'] else 'consume'
    self.update_phase(event_msg, phase=phase)
    await self.success_channel.put(event_msg)
```

Then the enricher plugin will add the necessary info for the audit log messages. 

Not sure if that helps at all, lmk if there's any outstanding confusion.

gordon_gcp/schema/parse.py
```diff
@@ -0,0 +1,82 @@
...
+To use:
+
+.. code-block:: pycon
+
+    >>> import json
+    >>> from gordon_gcp.schema import parse
+    >>> parser = parse.MessageParser()
+    >>> # using an example file
+    >>> exp = 'gordon_gcp/schema/examples/audit-log.last-operation.json'
+    >>> with open(exp, 'r') as r:
```
matt0bi: `with open(exp, 'r') as r:` -> `with open(exp, 'r') as **f**:`

econchick:
I addressed @descent3me comments - any other thoughts? from @descent3me or @spotify/alf ?

gordon_gcp/schema/parse.py
descent3me: These seem to be essentially static functions. What is the advantage of writing them as instance methods?

econchick: Only `_parse_event_msg` could be static, since `_parse_audit_log_msg` needs access to `self`. Though it could be a `classmethod` since it just needs access to the `ACTION_MAPPER`.

I'm not sure of the benefit for static/class methods versus instance methods. They are private methods, so only `parse` should be calling them. But maybe I'm missing something that you're thinking of?

descent3me: I was just thinking that it might be more complicated than it needs to be.  Why instantiate an object and pass around self, if it isn't needed? Or even why attach these functions to a object/class at all, rather than having `parse` as a module function?

econchick: The key driver for me across the board is to make everything extensible. This way, we don't restrict how users parse a message, i.e. maybe users want to access tags that are available in the audit log, or add a schema to parse against. I might be over-optimizing. @matt0bi any thoughts on the matter?

PR #7, opened 2018-03-29
Hotfix packaging

econchick:

If you try to install `gordon-gcp`, only the top-level modules are actually installed (the `__init__.py` and `exceptions.py`. It does not actually include `clients/*`, `plugins/*`, or `schema/*`.  This was hidden because `tox` runs tests against the actual directory when [it's supposed to run against the installed package via virtualenv](https://hynek.me/articles/testing-packaging).

Following the advice of @hynek via [his writeup](https://hynek.me/articles/testing-packaging), this PR moves the package into a `src` directory and fixes up `setup.py`. 

N/A

Once merged, I'll cut another dev release. I'll also update `gordon-janitor-gcp`. The `gordon` and `gordon-janitor` packages aren't (yet) affected because all modules are within top-level directory and not in subpackages.

**Release note**:

```release-note
Fixed packaging for distribution.
```

@matt0bi @spotify/alf


descent3me:
I'm not familiar with this approach, but I'm fine with trying it out. Can you give more detail on why it didn't work before - shouldn't the `recursive-include` in the manifest have included subdirectory modules?

econchick:
@descent3me I'm not sure how much more detail I can give beyond the linked writeup (which is pretty much exactly what happened here). Let me know if something from there is unclear.

With respect to why the recursive includes in the manifest didn't work: a manifest is required for source distributions (`sdist`); `python setup.py sdist` uses it for what to include in the `tar.gz` artifact. But that artifact is not actually installable; when `pip` is pointed to a tar, [it still has to build the package](https://packaging.python.org/tutorials/distributing-packages/#source-distributions). That's where the package definition within `setup.py:setup(...)` comes in, and where we were incorrect. 

It should also be noted that a manifest is only used for created source distributions, not wheels (which is preferred by `pip`).

setup.py
```diff
@@ -63,8 +64,8 @@ def install_requires():
 #####
 NAME = 'gordon-dns-gcp'
 PACKAGE_NAME = 'gordon_gcp'
-PACKAGES = ['gordon_gcp']
-META_PATH = os.path.join(PACKAGE_NAME, '__init__.py')
+PACKAGES = find_packages(where='src')
```
matt0bi: To make sure I understand this correctly: the old version of this was listing only the top-level package (gordon_gcp), which caused setuptools to include all the package _modules_, but it excluded any _sub-packages_ like `plugins`?

I've looked through the setuptools docs and the [pypa tutorial](https://packaging.python.org/tutorials/distributing-packages/#packages), but this part is very vague.

PR #11, opened 2018-04-06
Implement IEnricher interface

matt0bi:

This PR implements the IEnricherClient interface. It does not implement the Builder or the Spotify-specific parts of the code - those will be left for future PRs. 

N/A

This PR is rebased off #9 because it depends on certain parts of that PR to function, which is why this PR is in "RFC" state. If you'd like to give me feedback before #9 is merged, check out commit 5525260.

**Release note**:

```release-note
NONE
```

@spotify/alf

### DIFF
--- a/src/gordon_gcp/plugins/enricher.py
+++ b/src/gordon_gcp/plugins/enricher.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright 2017 Spotify AB
+# Copyright 2017-2018 Spotify AB
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -28,8 +28,14 @@ it will be placed into the appropriate channel, either the
 
 """
 
+import asyncio
+import logging
+
 import zope.interface
 from gordon import interfaces
+
+from gordon_gcp import exceptions
+from gordon_gcp.plugins import _utils
 
 
 __all__ = ('GCEEnricher',)
@@ -41,6 +47,8 @@ class GCEEnricher:
 
     Args:
         config (dict): configuration relevant to Compute Engine.
+        http_client (.AIOConnection): client for interacting with
+            the GCE API.
         success_channel (asyncio.Queue): a sink for successfully
             processed :interface:`interfaces.IEventMessages`.
         error_channel (asyncio.Queue): a sink for
@@ -49,11 +57,110 @@ class GCEEnricher:
     """
     phase = 'enrich'
 
-    def __init__(self, config, success_channel, error_channel):
+    def __init__(self, config, http_client, success_channel, error_channel):
         self.config = config
         self.success_channel = success_channel
         self.error_channel = error_channel
+        self._http_client = http_client
+        self._logger = logging.getLogger('')
 
-    def process(self):
-        # do something
-        pass  # pragma: no cover
+    async def _poll_for_instance_data(self, resource_name, msg_logger):
+        exception = None
+        backoff = 2
+        base_url = f'https://compute.googleapis.com/compute/v1/{resource_name}'
+
+        # Poll until instance data contains external IP information.
+        for attempt in range(1, self.config['retries'] + 1):
+            try:
+                msg_logger.debug(f'Attempt {attempt}: fetching {base_url}')
+                instance_data = await self._http_client.get_json(base_url)
+
+                network_interfaces = instance_data['networkInterfaces']
+                assert network_interfaces[0]['accessConfigs'][0]['natIP']
+
+                return instance_data
+            except exceptions.GCPHTTPError as e:
+                exception = e
+                break
+            except (KeyError, IndexError) as e:
+                exception = e
+                if attempt == self.config['retries']:
+                    continue
+                await asyncio.sleep(backoff)
+                backoff = backoff ** 2
+
+        msg = (f'Could not get external IP for {resource_name}: '
+               f'{exception.__class__.__name__}: {exception}')
+        raise exceptions.GCPGordonError(msg)
+
+    async def _create_A_rrecord(self, event_data, instance_data, msg_logger):
+        dns_zone = self.config['dns_zone']
+        hostname = event_data['resourceName'].split('/')[-1]
+        access_configs = instance_data['networkInterfaces'][0]['accessConfigs']
+        external_ip = access_configs[0]['natIP']
+
+        name = '.'.join([hostname, dns_zone])
+        msg_logger.debug(f'Creating A record: {name} -> {external_ip}')
+        return {
+            'name': name,
+            'rrdatas': [external_ip],
+            'type': 'A',
+            'ttl': self.config['default_ttl']
+        }
+
+    async def _create_rrecords(self, event_data, msg_logger):
+        instance_data = await self._poll_for_instance_data(
+            event_data['resourceName'], msg_logger)
+
+        a_record = await self._create_A_rrecord(event_data, instance_data,
+                                                msg_logger)
+        return [a_record]
+
+    async def process(self, event_message):
+        """
+        Enrich message with extra context and send it to the publisher.
+
+        When a message is successfully processed, it is passed to the
+        :obj:`self.success_channel`. However, if there is a problem
+        during processing, the message is passed to the
+        :obj:`self.error_channel`.
+
+        Args:
+            event_message (.GEventMessage): message requiring
+                additional information.
+        """
+        msg_logger = _utils.GEventMessageLogger(
+            self._logger, {'msg_id': event_message.msg_id})
+        data = event_message.data
+        try:
+            records = await self._create_rrecords(data, msg_logger)
+        except exceptions.GCPGordonError as e:
+            msg = ('Encountered error while enriching message, sending message'
+                   f' to error channel: {e}.')
+            msg_logger.warning(msg)
+            event_message.append_to_history(msg, self.phase)
+            await self.error_channel.put(event_message)
+            return
+        msg_logger.debug(f'Enriched with resource record(s): {records}')
+        event_message.data['resourceRecords'].extend(records)
+
+        added_records = event_message.data['resourceRecords']
+        msg = f'Enriched msg with {len(added_records)} resource record(s).'
+        event_message.append_to_history(msg, self.phase)
+
+        await self.update_phase(event_message, 'publish')
+        await self.success_channel.put(event_message)
+
+    async def update_phase(self, event_message, phase=None):
+        """
+        Update message's phase to next phase.
+
+        Args:
+            event_message (.GEventMessage): successfully processed message.
+            phase (str): next processing phase.
+        """
+        old_phase = event_message.phase
+        new_phase = phase or self.phase
+        event_message.phase = new_phase
+        msg = f'Updated phase from "{old_phase}" to "{new_phase}".'
+        event_message.append_to_history(msg, new_phase)
--- a/tests/unit/plugins/test_enricher.py
+++ b/tests/unit/plugins/test_enricher.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright 2017 Spotify AB
+# Copyright 2017-2018 Spotify AB
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -19,14 +19,34 @@ import asyncio
 import pytest  # NOQA
 from gordon import interfaces
 
+from gordon_gcp import exceptions
 from gordon_gcp.plugins import enricher
 
 
-def test_implements_interface():
+INSTANCE_NETWORK_DATA = {
+    'networkInterfaces': [{
+        'accessConfigs': [{
+            'natIP': '127.99.199.27'
+        }],
+    }]
+}
+
+
+@pytest.fixture
+def config(fake_keyfile):
+    return {
+        'keyfile': fake_keyfile,
+        'scopes': [],
+        'dns_zone': 'example.com.',
+        'default_ttl': 300,
+        'retries': 5
+    }
+
+
+def test_implements_interface(config):
     """GCEEnricher implements IEnricherClient"""
-    config = {'foo': 'bar'}
     success, error = asyncio.Queue(), asyncio.Queue()
-    client = enricher.GCEEnricher(config, success, error)
+    client = enricher.GCEEnricher(config, None, success, error)
 
     assert interfaces.IEnricherClient.providedBy(client)
     assert interfaces.IEnricherClient.implementedBy(enricher.GCEEnricher)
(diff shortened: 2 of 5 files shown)


tox.ini
```diff
@@ -10,7 +10,7 @@ setenv =
 deps = -rdev-requirements.txt
 commands =
     /usr/bin/find . -name "*.pyc" -delete
-    flake8 {posargs}
```
matt0bi: I removed this part because passing in any arguments to tox meant that they would either get passed to flake8 _or_  py.test. Since they have different arguments, this always led to one of them failing.

tests/unit/plugins/test_enricher.py
```diff
@@ -19,18 +19,132 @@
...
+    }]
+}
+
+
+@pytest.fixture
+def config(fake_keyfile):
+    return {
+        'keyfile': fake_keyfile,
+        'scopes': [],
+        'dns_zone': 'example.com',
```
econchick: trailing dot? `example.com.`?

tests/unit/plugins/test_enricher.py
```diff
@@ -19,18 +19,132 @@
...
     assert interfaces.IEnricherClient.providedBy(client)
     assert interfaces.IEnricherClient.implementedBy(enricher.GCEEnricher)
     assert config is client.config
     assert success is client.success_channel
     assert error is client.error_channel
     assert 'enrich' == client.phase
+
+
+@pytest.fixture
+def event_msg(mocker):
```
econchick: Maybe this could be moved to `conftest.py` since I also have a fixture like this in the event consumer plugin.

tests/unit/plugins/test_enricher.py
```diff
@@ -19,18 +19,132 @@
...
+            'timestamp': '2017-12-04T20:13:44.181Z',
+        }
+    pubsub_msg = mocker.Mock()
+    pubsub_msg._ack_id = '1:1'
+    return event_consumer.GEventMessage(
+        pubsub_msg, fake_data, enricher.GCEEnricher.phase)
+
+
+@pytest.fixture
+def channel_pair():
```
econchick: oo this too, to move into `conftest.py`

tests/unit/plugins/test_enricher.py
```diff
@@ -19,18 +19,132 @@
...
+    return asyncio.Queue(), asyncio.Queue()
+
+
+@pytest.fixture
+def mock_http_client(mocker, get_mock_coro):
+    get_json_mock, get_json_coro = get_mock_coro()
+    http_client = mocker.Mock()
+    mocker.patch.object(http_client, 'get_json', get_json_coro)
+    mocker.patch.object(http_client, '_get_json', get_json_mock)
```
econchick: Hmm this is a little confusing. Maybe instead of `_get_json`, `get_json.mock` or `get_json_mock` if not possible? I actually looked up if there was a `_get_json` method in the http client, lol.

tests/unit/plugins/test_enricher.py
```diff
@@ -19,18 +19,132 @@
...
+@pytest.fixture
+def channel_pair():
+    return asyncio.Queue(), asyncio.Queue()
+
+
+@pytest.fixture
+def mock_http_client(mocker, get_mock_coro):
+    get_json_mock, get_json_coro = get_mock_coro()
+    http_client = mocker.Mock()
+    mocker.patch.object(http_client, 'get_json', get_json_coro)
```
econchick: What's the diff between using `mocker.patch...` and `monkeypatch.setattr`? Is it just one less fixture to pass in?

tests/unit/plugins/test_enricher.py
```diff
@@ -19,18 +19,132 @@
...
+    gce_enricher = enricher.GCEEnricher(config, mock_http_client, success_chnl,
+                                        error_chnl)
+
+    await gce_enricher.process(event_msg)
+
+    event_message = await success_chnl.get()
+
+    assert 'publish' == event_message.phase
+    expected_history_msg = 'Enriched msg 1 with resourceRecords'
+    assert expected_history_msg == event_message.history_log[0]['message']
```
econchick: maybe also assert `'enrich' == event_message.history_log[0]['plugin']`?

tests/unit/plugins/test_enricher.py
```diff
@@ -19,18 +19,132 @@
...
+                                 mock_async_sleep, mock_http_client, caplog,
+                                 network_data, sleep_calls, logs_logged):
+    success_chnl, error_chnl = channel_pair
+    mock_http_client._get_json.side_effect = network_data
+    gce_enricher = enricher.GCEEnricher(config, mock_http_client, success_chnl,
+                                        error_chnl)
+
+    await gce_enricher.process(event_msg)
+
+    event_message = await success_chnl.get()
```
econchick: Naming is a bit confusing with `event_msg` fixture. Maybe something like `ret_event_msg`? just something to clearly differentiate.

tests/unit/plugins/test_enricher.py
```diff
@@ -19,18 +19,132 @@
...
+    return sleep_mock
+
+
+@pytest.mark.parametrize('network_data,sleep_calls,logs_logged', [
+    ([INSTANCE_NETWORK_DATA], 0, 3),
+    ([{}, {}, INSTANCE_NETWORK_DATA], 2, 5)])
+@pytest.mark.asyncio
+async def test_process_event_msg(config, event_msg, channel_pair,
+                                 mock_async_sleep, mock_http_client, caplog,
+                                 network_data, sleep_calls, logs_logged):
```
econchick: docstring?

tests/unit/plugins/test_enricher.py
```diff
@@ -19,18 +19,132 @@
...
+
+
+@pytest.mark.parametrize('response,sleep_calls,logs_logged', [
+    (exceptions.GCPHTTPError('404 error'), 0, 2),
+    ([{}] * 5, 4, 6)])
+@pytest.mark.asyncio
+async def test_process_event_msg_failures(config, event_msg, channel_pair,
+                                          mock_async_sleep, mock_http_client,
+                                          caplog, response, sleep_calls,
+                                          logs_logged):
```
econchick: docstring?

tests/unit/plugins/test_enricher.py
```diff
@@ -19,18 +19,132 @@
...
+
+    await gce_enricher.process(event_msg)
+
+    event_message = await error_chnl.get()
+
+    assert 'enrich' == event_message.phase
+    assert event_msg == event_message
+    assert sleep_calls == mock_async_sleep.call_count
+    assert logs_logged == len(caplog.records)
+    assert 0 == len(event_message.history_log)
```
econchick: Hmm maybe there should be a log entry for the message history about this. What do you think?

tests/unit/plugins/test_enricher.py
```diff
@@ -19,18 +19,132 @@
...
+                                          logs_logged):
+    success_chnl, error_chnl = channel_pair
+    mock_http_client._get_json.side_effect = response
+    gce_enricher = enricher.GCEEnricher(config, mock_http_client, success_chnl,
+                                        error_chnl)
+
+    await gce_enricher.process(event_msg)
+
+    event_message = await error_chnl.get()
+
```
econchick: What do you think of adding an `assert success_chnl.empty()`?

src/gordon_gcp/plugins/enricher.py
```diff
@@ -49,11 +56,87 @@ class GCEEnricher:
...
+            a_record = await self._create_A_rrecord(data)
+        except exceptions.GCPGordonError as e:
+            msg = ('Encountered error while enriching message, sending message'
+                   f' to error channel: {e}')
+            logging.warning(msg)
+            await self.error_channel.put(event_message)
+            return
+        logging.debug(f'Created resource record {a_record}')
+        event_message.data['resourceRecords'].append(a_record)
+        msg = f'Enriched msg {event_message.msg_id} with resourceRecords'
```
econchick: Hmm - I wonder if could this history log message be more helpful. You shouldn't need the message's ID since the history log attached to the object itself. I wonder if it'd be helpful to include something like `Enriched msg with {len(added_a_records)} A records.` (and maybe so we can extend it to `{len(added_cname_records)} CNAME records` and TXT etc.)

Also - would you mind adding a period at the end of your (event) log messages? I think I tried to do that consistently w/ my patch. Total nit, but I feel like I'd go crazy with inconsistency.

src/gordon_gcp/plugins/enricher.py
```diff
@@ -49,11 +56,87 @@ class GCEEnricher:
...
+        data = event_message.data
+        try:
+            a_record = await self._create_A_rrecord(data)
+        except exceptions.GCPGordonError as e:
+            msg = ('Encountered error while enriching message, sending message'
+                   f' to error channel: {e}')
+            logging.warning(msg)
+            await self.error_channel.put(event_message)
+            return
+        logging.debug(f'Created resource record {a_record}')
```
econchick: When you rebase my stuff, could you use the message logger so we get the message ID?

src/gordon_gcp/plugins/enricher.py
```diff
@@ -49,11 +56,87 @@ class GCEEnricher:
...
+
+        msg = f'Could not get external IP for {resource_name}: {exception}'
+        raise exceptions.GCPGordonError(msg)
+
+    async def _create_A_rrecord(self, data):
+        external_ip = await self._poll_for_ext_ip(data['resourceName'])
+        dns_zone = self.config['dns_zone']
+        hostname = data['resourceName'].split('/')[-1]
+
+        name = '.'.join([hostname, dns_zone])
```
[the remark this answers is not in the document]
matt0bi: Good catch - I'll add this to the builder's validation flow in #12 .

src/gordon_gcp/plugins/enricher.py
```diff
@@ -49,11 +56,87 @@ class GCEEnricher:
...
+        logging.debug(f'Creating A record: {name} -> {external_ip}')
+        return {
+            'name': name,
+            'rrdatas': [external_ip],
+            'type': 'A'
```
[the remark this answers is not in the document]
matt0bi: Hm, I set it up so that it's part of the enricher config/workflow - could you take a look and see if it makes sense this way? This would keep event message changes to just one plugin. What do you think?

tests/unit/plugins/test_enricher.py
matt0bi: Thanks for pointing it out. Changed it to `get_json_mock`.

> I actually looked up if there was a _get_json method in the http client, lol.

GOT YOU! :grin:

tests/unit/plugins/test_enricher.py
matt0bi: Comparing [monkeypatch](https://docs.pytest.org/en/latest/reference.html#_pytest.monkeypatch.MonkeyPatch.setattr) and [mocker.patch.object](https://docs.python.org/3/library/unittest.mock.html#patch-object) there doesn't seem to be a difference in this case except the "one less fixture". I think I used it because I was using `mocker` everywhere and forgot about `monkeypatch`.

tests/unit/plugins/test_enricher.py
matt0bi: ಠಿ_ಠಿ, good point!

src/gordon_gcp/plugins/enricher.py
```diff
@@ -49,11 +56,87 @@ class GCEEnricher:
...
+            await self.error_channel.put(event_message)
+            return
+        logging.debug(f'Created resource record {a_record}')
+        event_message.data['resourceRecords'].append(a_record)
+        msg = f'Enriched msg {event_message.msg_id} with resourceRecords'
+        event_message.append_to_history(msg, self.phase)
+        await self.update_phase(event_message, 'publish')
+        await self.success_channel.put(event_message)
+
+    async def update_phase(self, event_message, phase):
```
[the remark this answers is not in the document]
matt0bi: > Maybe this should live on the message itself? since it's acting on the message?

Talked about about this IRL: this will be moved to the msg when working on core routing logic.

matt0bi:
Hey @econchick, I've addressed your feedback, could PTAL? Also, some of your feedback also applies to #12, so I'll go over that on next occasion and apply it there.

src/gordon_gcp/plugins/enricher.py
```diff
@@ -49,11 +57,96 @@ class GCEEnricher:
...
+        except exceptions.GCPGordonError as e:
+            msg = ('Encountered error while enriching message, sending message'
+                   f' to error channel: {e}.')
+            msg_logger.warning(msg)
+            event_message.append_to_history(msg, self.phase)
+            await self.error_channel.put(event_message)
+            return
+        msg_logger.debug(f'Created resource record {a_record}')
+        event_message.data['resourceRecords'].append(a_record)
+        added_a_records = event_message.data['resourceRecords']
```
econchick: So, hm, how would someone (us, too) extend this to process more record types? It doesn't seem like one would be able to without rewriting all of `process`. Maybe, `process` could iterate over some registration of record types that are mapped to their respective methods. Something like (I have no idea if this would work, and this is just the first approach I thought of, there may be better ones):

```python
# original
class GCEEnricher:
    RECORD_HANDLERS = {
        'A': '_create_A_rrecord',  # string name of method
    }
    async def process(self, msg):
        # prep stuff
        for rtype, method in RECORD_HANDLERS.items():
            records = await getattr(self, method)(data, msg_logger)
        # other stuff

# extended
class MyEnricher(GCEEnricher):
    RECORD_HANDLERS.update({ 
        'CNAME': '_create_cname_rrecord',
    }
```

src/gordon_gcp/plugins/enricher.py
```diff
@@ -49,11 +57,96 @@ class GCEEnricher:
...
+    def __init__(self, config, http_client, success_channel, error_channel):
         self.config = config
         self.success_channel = success_channel
         self.error_channel = error_channel
+        self._http_client = http_client
+        self._logger = logging.getLogger('')
+
+    async def _poll_for_ext_ip(self, resource_name, msg_logger):
+        exception = None
+        backoff = 2
```
econchick: Hmm we should put backoff in the http client eventually 🤔

matt0bi: Created a ticket off this.

src/gordon_gcp/plugins/enricher.py
matt0bi: Thanks for pointing this out. I went with the approach we're using in the Janitor code eg. overwritable methods. Users that want more records can override the `_create_rrecords` method to call other, new methods for creating specific records. I'm not 100% sure this is the most comfortable way, but it seems the simplest.

src/gordon_gcp/plugins/enricher.py
```diff
@@ -49,11 +57,110 @@ class GCEEnricher:
...
+        base_url = f'https://compute.googleapis.com/compute/v1/{resource_name}'
+
+        # Poll until instance data contains external IP information.
+        for attempt in range(1, self.config['retries'] + 1):
+            try:
+                msg_logger.debug(f'Attempt {attempt}: fetching {base_url}')
+                instance_data = await self._http_client.get_json(base_url)
+
+                a_confs = instance_data['networkInterfaces'][0]['accessConfigs']
```
matt0bi: This piece of code simply checks for the existence of a nested key. I think assigning this variable is the lesser evil compared to two other alternatives I thought of: adding an `if` block, which adds yet one more level of indentation, or pulling this out into its own tiny function. Any thoughts/advice?

src/gordon_gcp/plugins/enricher.py
```diff
@@ -49,11 +57,96 @@ class GCEEnricher:
...
+        When a message is successfully processed, it is passed to the
+        :obj:`self.success_channel`. However, if there is a problem
+        during processing, the message is passed to the
+        :obj:`self.error_channel`.
+
+        Args:
+            event_message (.GEventMessage): message requiring
+                additional information.
+        """
+        msg_logger = _utils.GEventMessageLogger(
```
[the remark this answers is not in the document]
matt0bi: I think there is - do you want to talk about this IRL today?

src/gordon_gcp/plugins/enricher.py
econchick: Another approach could be an `assert` statement:

```python
assert instance_data['networkInterfaces'][0]['accessConfigs'][0]['natIP'], 'some error message'
```

(maybe shorten that dict lookup somehow)

I don't necessarily have strong opinions either way.

src/gordon_gcp/plugins/enricher.py
```diff
@@ -49,11 +57,110 @@ class GCEEnricher:
...
-        # do something
-        pass  # pragma: no cover
+        Args:
+            event_message (.GEventMessage): successfully processed message.
+            phase (str): next processing phase.
+        """
+        old_phase = event_message.phase
+        new_phase = phase or self.phase
+        event_message.phase = new_phase
+        msg = f'Updated phase from "{old_phase}" to "{phase}".'
```
econchick: Hm shouldn't this be `new_phase` instead of `phase`?

src/gordon_gcp/plugins/enricher.py
matt0bi: Ah, thanks. Makes the intent more explicit :+1:

matt0bi:
Thanks @econchick, I addressed the two pieces of feedback! Will merge this in either today or Monday.

PR #12, opened 2018-04-11
[RFC] Add GCEEnricherBuilder class

matt0bi:

This PR adds a Builder class for the GCEEnricher class and wires it up to `setup.py`. Additionally, it deduplicates some `test_init.py` fixtures by moving them to `conftest.py`.

N/A

This PR is based off of #9 and #11 - so only efe4157 and 315a681 should be reviewed.
**Release note**:

```release-note
NONE
```

@spotify/alf


tests/unit/plugins/test_init.py
```diff
@@ -0,0 +1,258 @@
...
+                                          auth_client):
+    """Raise when 'dns_zone' config value doesn't end with root zone."""
+    mocker.patch('gordon_gcp.plugins.enricher.http.AIOConnection')
+    success_chnl, error_chnl = asyncio.Queue(), asyncio.Queue()
+    enricher_config['dns_zone'] = 'example.com'
+
+    with pytest.raises(exceptions.GCPConfigError) as e:
+        plugins.get_enricher(enricher_config, success_chnl, error_chnl)
+
+    exc_msg = 'A dns zone must be a FQDN and end with the root zone \("."\).'
```
matt0bi: TIL that `e.match` uses regexp matching and not string comparison so regexp-specific characters like `(` have to be escaped.

matt0bi:
Hey @econchick, I've rebased this PR from the `add-event-consumer` branch and applied some of the feedback from #11 (eg. tests have docstrings now) so this one's ready for review/comments.

PR #13, opened 2018-04-27
RFC: Alf 5213 http retry

ocsw:

Fixes #

**Release note**:

```release-note

```

@spotify/alf no tests yet, and I'm going to add more logging, but this is the basic idea.

### DIFF
--- a/src/gordon_gcp/clients/http.py
+++ b/src/gordon_gcp/clients/http.py
@@ -19,10 +19,6 @@ Module to interact with Google APIs via asynchronous HTTP calls.
 product-specific API clients as it handles Google authentication and
 automatic refresh of tokens.
 
-.. todo::
-
-    Include that it also handles retries once implemented.
-
 To use:
 
 .. code-block:: python
@@ -39,7 +35,7 @@ To use:
 
 import datetime
 import functools
-import http.client
+import http.client.UNAUTHORIZED
 import json
 import logging
 
@@ -51,9 +47,6 @@ from gordon_gcp.clients import _utils
 
 
 __all__ = ('AIOConnection',)
-
-MAX_REFRESH_ATTEMPTS = 2
-REFRESH_STATUS_CODES = (http.client.UNAUTHORIZED,)
 
 
 class AIOConnection:
@@ -86,9 +79,9 @@ class AIOConnection:
             await self._auth_client.refresh_token()
 
     async def request(self, method, url, params=None, body=None, headers=None,
-                      retry_type: "none", retries=1, timeout=None,  # TODO
+                      retry_type="none", tries=1, timeout=None,
                       retry_predicate=None, predicate_args=None,
-                      predicate_kwargs=None):
+                      predicate_kwargs=None, extra_401_retry=True):
         """Wrapper for _request() that handles retries and backoffs.
 
         Args:
@@ -102,29 +95,36 @@ class AIOConnection:
                 request. Headers pass through to the request will
                 include :attr:`DEFAULT_REQUEST_HEADERS`.
             retry_type (str): (optional) The type of retries to use: "none",
-                "simple", (retry up to retries/timeout), "no4xx" (don't retry
-                on 4xx, otherwise up to retries/timeout), or "custom".  If
+                "simple", (retry up to tries/timeout), "no4xx" (don't retry
+                on 4xx, otherwise up to tries/timeout), or "custom".  If
                 "custom", uses the retry predicate specified in the other
-                parameters.
-            retries TODO
-            timeout TODO
+                parameters.  See also extra_401_retry.
+            tries (int): (optional) Total maximum number of times to try the
+                request, including the first (i.e. 1 == no retries).  The number
+                of actual tries is subject to retry_type and timeout as well as
+                this value.
+            timeout (int): (optional) Total maximum seconds to keep trying the
+                request (None == unlimited).  The number of actual tries is subject to retry_type and tries as well as this value.
             retry_predicate (callable): (optional) Ignored unless retry_type is
                 "custom".  Last argument must be the response from _request().
-                Must return True for "continue retrying (subject to retries and
+                Must return True for "continue retrying (subject to tries and
                 timeout)" or False for "stop retrying".  If the function takes
                 any other arguments, their values must be supplied in
                 predicate_args and/or predicate_kwargs.
             predicate_args (list): (optional) See retry_predicate.
             predicate_kwargs (dict): (optional) See retry_predicate.
+            extra_401_retry (bool): (optional) If True, run one more request
+                attempt on a 401 (unauthorized), regardless of the value of
+                retry_type.  This makes sure we've tried generating a new token.
         Returns:
             (str) HTTP response body.
         Raises:
             :exc:`.GCPHTTPError`: If a response code >=400 was received and
-                there are no more retries (or no retrying is specified).
+                there are no more tries (or no retrying is specified).
         """
         retry_predicates = {
             "none": lambda resp: False,  # never retry
-            "simple": lambda resp: True,  # always retry up to retries/timeout
+            "simple": lambda resp: True,  # always retry up to tries/timeout
             # don't retry on 4xx
             "no4xx": lambda resp: resp.status < 400 or resp.status >=500,
             "custom": retry_predicate,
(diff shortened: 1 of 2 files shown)


ocsw:
Should we move token regeneration into the wrapper?

src/gordon_gcp/clients/http.py
```diff
@@ -38,21 +34,20 @@
 """
 
 import datetime
-import http.client
 import json
 import logging
+from functools import partial
+from http import HTTPStatus
```
econchick: These two imports goes against [our following](https://github.com/spotify/gordon-gcp/blob/develop/CONTRIBUTING.rst#code) of Google's style guide.

src/gordon_gcp/clients/http.py
```diff
@@ -83,9 +78,11 @@ def __init__(self, auth_client=None, session=None):
...
             await self._auth_client.refresh_token()
 
-    async def request(self, method, url, params=None, body=None,
-                      headers=None, **kwargs):
-        """Make an asynchronous HTTP request.
+    async def request(self, method, url, params=None, body=None, headers=None,
+                      retry_type="none", tries=1, timeout=None,
+                      retry_predicate=None, predicate_args=None,
+                      predicate_kwargs=None, extra_401_retry=True):
+        """Wrapper for _request() that handles retries and backoffs.
```
econchick: This sort of docstring isn't terribly helpful. Even if it's a wrapper for something, it wraps a (semi) private method, whose docstring can't be assumed would be read (or even written) since it's not a part of the public api. I think the old docstring still makes sense since it's the public API to make an http request.

src/gordon_gcp/clients/http.py
```diff
@@ -83,9 +78,11 @@ def __init__(self, auth_client=None, session=None):
         if not is_valid:
             await self._auth_client.refresh_token()
 
-    async def request(self, method, url, params=None, body=None,
-                      headers=None, **kwargs):
-        """Make an asynchronous HTTP request.
+    async def request(self, method, url, params=None, body=None, headers=None,
+                      retry_type="none", tries=1, timeout=None,
+                      retry_predicate=None, predicate_args=None,
+                      predicate_kwargs=None, extra_401_retry=True):
```
econchick: I think I may understand the motivation for this overall approach - multiple keyword arguments for the ability to address many use cases/configurability. Do I understand correctly?

I fear this might be too much at once, and I'd much prefer to see the most common case addressed first, then we add what we need when we need it.

May I ask: is there a reason you chose to not use decorators? I suspect maybe because decorators are evaluated at runtime, making it more difficult to configure. I think it might be possible ([docs](https://github.com/litl/backoff#runtime-configuration)) since plugins are loaded and config [should be available](https://github.com/spotify/gordon/blob/develop/gordon/plugins_loader.py#L67-L91). Perhaps you already tried that though.

src/gordon_gcp/clients/http.py
```diff
@@ -95,15 +92,96 @@ def __init__(self, auth_client=None, session=None):
...
+                         " -> Failed predicate, trying again "
+                         f"({details.tries} tries so far).")
+
+        def giveup_logger(details):
+            """See https://github.com/litl/backoff#event-handlers."""
+            logging.info(_utils.REQ_LOG_FMT.format(method, url) +
+                         " -> Failed predicate, but no more retries "
+                         f"({details.tries} tries so far).")
+
+        resp = await backoff.on_predicate(
```
econchick: What's the reasoning for backoff on predicate, rather than on exception? is it configurability?

src/gordon_gcp/clients/http.py
```diff
@@ -95,15 +92,96 @@ def __init__(self, auth_client=None, session=None):
...
+        actual_predicate = retry_predicates[retry_type]
+        if predicate_args is None:
+            predicate_args = []
+        if predicate_kwargs is None:
+            predicate_kwargs = {}
+        predicate_partial = partial(
+            actual_predicate, *predicate_args, **predicate_kwargs
+        )
 
+        def backoff_logger(details):
```
econchick: I'm guilty of using inner functions, but they make it very difficult to test so I try to only use them when necessary. Something like these two loggers, I'm not sure are necessary. If backoff's [logger](https://github.com/litl/backoff#logging-configuration) is missing something, we should adjust it in that logger itself (which could go in the initialization of this http client, or even setup on the module level/global).

src/gordon_gcp/clients/http.py
```diff
@@ -19,10 +19,6 @@
 product-specific API clients as it handles Google authentication and
 automatic refresh of tokens.
 
-.. todo::
-
-    Include that it also handles retries once implemented.
```
econchick: Could you update the module-level docs to reflect the new functionality?

src/gordon_gcp/clients/http.py
```diff
@@ -95,15 +92,96 @@ def __init__(self, auth_client=None, session=None):
...
+                         " -> Failed predicate, but no more retries "
+                         f"({details.tries} tries so far).")
+
+        resp = await backoff.on_predicate(
+            backoff.expo, max_tries=tries, max_time=timeout,
+            predicate=predicate_partial, on_backoff=backoff_logger,
+            on_giveup=giveup_logger,
+        )(self._request)(method, url, params, body, headers)
+        # one extra try on a 401 to make sure we've refreshed the token
+        if resp.status == HTTPStatus.UNAUTHORIZED and extra_401_retry:
```
econchick: I was following [google's approach](https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/2857265d946875dca97eec706e6eb9ed1412d677/google/auth/transport/__init__.py#L32) in having a tuple of http codes to refresh tokens (and as module-level constants); it gave me the impression that Google may add more codes in the future.

Also, I think the `extra_401_retry` might be serving a very tiny use case (is there a particular one you had in mind?) that doesn't need to be addressed in the first pass.

src/gordon_gcp/clients/http.py
```diff
@@ -95,15 +92,96 @@ def __init__(self, auth_client=None, session=None):
...
+            logging.info(_utils.REQ_LOG_FMT.format(method, url) +
+                         " -> Unauthorized.  Trying one more time with new "
+                         "token.")
```
econchick: What's the reasoning for changing the log message?

econchick: Also, the style guide uses single quotes for strings.

src/gordon_gcp/clients/http.py
```diff
@@ -95,15 +92,96 @@ def __init__(self, auth_client=None, session=None):
...
+                retry_type is "custom".  Must return True for "continue retrying
+                (subject to tries and timeout)" or False for "stop retrying".
+                Last argument must be the response from _request().  If the
+                function takes any other arguments, their values must be
+                supplied in predicate_args and/or predicate_kwargs.
+            predicate_args (list): (optional) See retry_predicate.
+            predicate_kwargs (dict): (optional) See retry_predicate.
+            extra_401_retry (bool): (optional) If True, run one more request
+                attempt on a 401 (unauthorized), regardless of the value of
+                retry_type.  This makes sure we've tried generating a new token.
```
econchick: If we can't make decorators work at all, then I suggest lumping all `backoff`-related keyword arguments in one, something like `backoff_kwargs={...}`. Then creating a helper method to parse all that is needed. I completely understand the preference to be explicit, but this method is a bit overloaded and deceptively complex, and the keyword args defined are not necessarily obviously applicable to backoff functionality. Having it all in one kwarg would remove the helpful docstrings, so for that, I'd suggest writing [prose documentation](https://github.com/spotify/gordon-gcp/blob/develop/docs/clients.rst#asynchronous-gcp-http-client) and referring to that in this docstring.

src/gordon_gcp/clients/http.py
```diff
@@ -95,15 +92,96 @@ def __init__(self, auth_client=None, session=None):
...
         Returns:
             (str) HTTP response body.
         Raises:
-            :exc:`.GCPHTTPError`: if any exception occurred.
+            :exc:`.GCPHTTPError`: If a response code >=400 was received and
+                there are no more tries (or no retrying is specified).
         """
-        refresh_attempt = kwargs.pop('cred_refresh_attempt', 0)
+        retry_predicates = {
+            "none": lambda resp: False,  # never retry
```
econchick: If we end up going down this route, I think it's more pythonic to do something like this:

```py
# retry_type=None in method kwargs
predicate = lambda resp: False  # set default
retry_predicates = {
    'all': lambda resp: resp.status >= 400,
    'server_errors': lambda resp: resp.status >= 500,
    'custom': retry_predicate
}
try:
    predicate = retry_predicate[retry_type]
except KeyError:
    pass  # end up using the default
```

src/gordon_gcp/clients/http.py
```diff
@@ -95,15 +92,96 @@ def __init__(self, auth_client=None, session=None):
...
+            # don't retry on 4xx
+            "no4xx": lambda resp: resp.status >= 500,
+            "custom": retry_predicate,
+        }
+        actual_predicate = retry_predicates[retry_type]
+        if predicate_args is None:
+            predicate_args = []
+        if predicate_kwargs is None:
+            predicate_kwargs = {}
+        predicate_partial = partial(
```
econchick: (again, if we end up going down this route) I think the user of this method should construct their own partial and supply this, since it's not relevant to the non-custom predicates.

src/gordon_gcp/clients/http.py
matt0bi: I think this idea is pretty good. I know some other libraries that do a lot of args/kwargs (mock, async-ssh, paramiko) and using them even for simple cases has me looking up the docs almost every time.

src/gordon_gcp/clients/http.py
```diff
@@ -83,9 +85,11 @@ def __init__(self, auth_client=None, session=None):
         if not is_valid:
             await self._auth_client.refresh_token()
 
-    async def request(self, method, url, params=None, body=None,
-                      headers=None, **kwargs):
-        """Make an asynchronous HTTP request.
+    async def request(self, method, url, params=None, body=None, headers=None,
+                      retry_type: "none", retries=1, timeout=None,  # TODO
```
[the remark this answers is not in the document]
ocsw: Already fixed.

src/gordon_gcp/clients/http.py
ocsw: Will fix.

src/gordon_gcp/clients/http.py
ocsw: Will fix.

src/gordon_gcp/clients/http.py
ocsw: Yup.

src/gordon_gcp/clients/http.py
ocsw: I'm fine with using backoff's logger if you are.  Wasn't sure it was worth having another logger to configure.

src/gordon_gcp/clients/http.py
ocsw: Will fix.

src/gordon_gcp/clients/http.py
ocsw: I can put the tuple back.

The 401 was in the original version.  I wasn't sure how we wanted to handle it in the new version.  Especially because if we use 'no4xx', do we still want to try once in case the problem was the token?  There's no way to change the retries after the fact based on the response.

src/gordon_gcp/clients/http.py
ocsw: Just making it more consistent with the others.  I can change it back if you hate it.

Am I missing something in the style guide?  What I saw says 'use either one as long as you're consistent' - which this isn't, so I'll change it, but I want to make sure I understand.

src/gordon_gcp/clients/http.py
ocsw: Makes sense I guess, although I need to figure out what that looks like.  Also not sure I got the message of your prose documentation link?

src/gordon_gcp/clients/http.py
ocsw: That would silently swallow a bad value.  I wanted to cause a KeyError on bad value.

ocsw: I can note that explicitly if that changes anything.

src/gordon_gcp/clients/http.py
ocsw: I can do that.  I just wasn't sure if it's putting too much work on the caller, vs just centralizing the partial() call.

src/gordon_gcp/clients/http.py
ocsw: You're correct about the motivation.  I agree it might be too much, but it seemed like a small amount of code to add to gain more functionality.  Maybe the calculus changes with your suggestion below of using predicate_kwargs?

The decorator thing I don't think can work unless we move _request() into request() as an inner function, but I'm open to discussing it.  Certainly easier to read in that case.

ocsw:
this will be in v2, closing until then

PR #16, opened 2018-05-08
Typos, formatting, comments in enricher

ocsw:
Fix a typo, add a couple of clarifying comments, and fix formatting of a function call.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/src/gordon_gcp/plugins/enricher.py
+++ b/src/gordon_gcp/plugins/enricher.py
@@ -65,6 +65,7 @@ class GCEEnricherBuilder:
 
     def _validate_config(self):
         # req keys: dns_zone, keyfile
+        # opt keys: retries
         errors = []
         if not self.config.get('keyfile'):
             msg = ('The path to a Service Account JSON keyfile is required to '
@@ -76,7 +77,7 @@ class GCEEnricherBuilder:
             errors.append(msg)
 
         if not self.config.get('dns_zone', '').endswith('.'):
-            msg = 'A dns zone must be a FQDN and end with the root zone (".").'
+            msg = 'A dns zone must be an FQDN and end with the root zone (".").'
             errors.append(msg)
 
         if errors:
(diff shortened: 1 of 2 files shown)


src/gordon_gcp/plugins/enricher.py
```diff
@@ -76,7 +77,7 @@ def _validate_config(self):
             errors.append(msg)
 
         if not self.config.get('dns_zone', '').endswith('.'):
-            msg = 'A dns zone must be a FQDN and end with the root zone (".").'
+            msg = 'A dns zone must be an FQDN and end with the root zone (".").'
             errors.append(msg)
```
descent3me: Actually, it's unrelated to typos, but I don't see the point in fixing typos for something that should change. If you wanted to make a new commit to just drop this whole if-block, and maybe reword the msg in the one above, that would be better, so that it would match the format expected by the janitor https://github.com/spotify/gordon-janitor-gcp/blob/develop/src/gordon_janitor_gcp/plugins/authority.py#L92

src/gordon_gcp/plugins/enricher.py
```diff
@@ -89,6 +90,7 @@ def _validate_config(self):
             self.config['retries'] = 5
 
     def _init_auth(self):
+        # req keys: scopes, keyfile
```
descent3me: I'm not sure this makes sense to include here, since the validation should already have happened at this point and we know `keyfile` is in the config. `scopes` is not required https://github.com/spotify/gordon-gcp/blob/develop/src/gordon_gcp/clients/auth.py#L80

src/gordon_gcp/plugins/enricher.py
ocsw: If I understand you correctly, the janitor one seems like it's missing the FQDN test.  Should we change that side?

descent3me: It was intentionally removed because none of the DNS provider APIs that I checked (Google, DYN, NS1, Route53)  expect a zone identifier in that format

descent3me: I retract the above comment.  I now (again) think it might be a good idea to use the absolute zone name, regardless of what the API's expect, since I think it makes more sense to us logically while reasoning about the system.  I will update `gordon-janitor-gcp`

descent3me:
I would update the commit message with a verb like "Fix", but otherwise +1

PR #17, opened 2018-05-08
Make enricher easily expandable

ocsw:
Makes it easy to write the internal bits that add more resource records.

Fixes #

**Release note**:

```release-note

```


src/gordon_gcp/plugins/enricher.py
```diff
@@ -156,16 +157,14 @@ def __init__(self, config, http_client, success_channel, error_channel):
                f'{exception.__class__.__name__}: {exception}')
         raise exceptions.GCPGordonError(msg)
 
-    async def _create_A_rrecord(self, event_data, instance_data, msg_logger):
-        dns_zone = self.config['dns_zone']
-        hostname = event_data['resourceName'].split('/')[-1]
+    async def _create_A_rrecord(self, fqdn, event_data, instance_data,
+                                msg_logger):
```
descent3me: I guess event data isn't needed anymore?

descent3me: This didn't change with this PR, but I also don't understand why this method is async

src/gordon_gcp/plugins/enricher.py
```diff
@@ -175,9 +174,16 @@ def __init__(self, config, http_client, success_channel, error_channel):
...
 
-        a_record = await self._create_A_rrecord(event_data, instance_data,
-                                                msg_logger)
-        return [a_record]
+        dns_zone = self.config['dns_zone']
+        hostname = event_data['resourceName'].split('/')[-1]
+        fqdn = '.'.join([hostname, dns_zone])
+
+        rrecords = []
+        for rrecord_creator in self._rrecord_creators:
```
descent3me: This might be a limited-scope optimization since it only allows a subclass to create more records within the same zone. If I were writing a subclass, I would be more inclined to override this whole `_create_rrecords` method, so I could have more flexibility both to use more zones and for the way in which I could design the class.

src/gordon_gcp/plugins/enricher.py
ocsw: Well, it might make sense to make all of the record creators async in case they need to use aiohttp or something.  Can't make only one async.

src/gordon_gcp/plugins/enricher.py
```diff
@@ -156,28 +156,50 @@ def __init__(self, config, http_client, success_channel, error_channel):
...
+    async def _get_full_data(self, event_data, msg_logger):
         instance_data = await self._poll_for_instance_data(
             event_data['resourceName'], msg_logger)
 
-        a_record = await self._create_A_rrecord(event_data, instance_data,
-                                                msg_logger)
-        return [a_record]
+        dns_zone = self.config['dns_zone']
+        hostname = event_data['resourceName'].split('/')[-1]
+        fqdn = '.'.join([hostname, dns_zone])
```
descent3me: Does not include trailing dot.

src/gordon_gcp/plugins/enricher.py
```diff
@@ -156,28 +156,50 @@ def __init__(self, config, http_client, success_channel, error_channel):
...
+
+        interfaces = instance_data['networkInterfaces']
+        external_ip = interfaces[0]['accessConfigs'][0]['natIP']
```
descent3me: I don't think this is correct.

src/gordon_gcp/plugins/enricher.py
```diff
@@ -156,28 +156,50 @@ def __init__(self, config, http_client, success_channel, error_channel):
...
+        internal_ip = None
+        if len(interfaces) > 1:
+            internal_ip = interfaces[1]['accessConfigs'][0]['natIP']
```
descent3me: Nor this one.

src/gordon_gcp/plugins/enricher.py
```diff
@@ -156,28 +156,50 @@ def __init__(self, config, http_client, success_channel, error_channel):
...
+
+        default_ttl = self.config['default_ttl']
```
descent3me: I think this didn't change with this PR, but this can cause a runtime KeyError, since it is required but not validated.

src/gordon_gcp/plugins/enricher.py
```diff
@@ -156,28 +156,50 @@ def __init__(self, config, http_client, success_channel, error_channel):
...
-            'name': name,
+            'name': fqdn,
             'rrdatas': [external_ip],
             'type': 'A',
-            'ttl': self.config['default_ttl']
+            'ttl': default_ttl
         }
 
-    async def _create_rrecords(self, event_data, msg_logger):
+    async def _get_full_data(self, event_data, msg_logger):
```
descent3me: If we're adding returning an arbitrary dict that we would expect a subclass to use, we should definitely add a docstring describing what it contains.

src/gordon_gcp/plugins/enricher.py
```diff
@@ -152,32 +161,49 @@ def __init__(self, config, http_client, success_channel, error_channel):
...
+        return interfaces[0]['accessConfigs'][0]['natIP']
+
+    def _get_internal_ip(self, instance_data, msg_logger):
+        # pulled out for easy subclassing
+        interfaces = instance_data['networkInterfaces']
+        return interfaces[0]['networkIP']
+
+    async def _create_rrecords(self, event_data, instance_data, msg_logger):
+        # async in case anything gets added in a subclass that needs to be async
+        fqdn = self._get_fqdn(event_data['resourceName'], msg_logger)
```
descent3me: Why not pull name from `instance_data['name']` instead?  Wherever we get the unqualified hostname from, I think passing it in instead could simplify the `_get_fqdn` function, but if we use instance data as the source we could erase the need for the for the `_get_hostname` method as well.

I realize this didn't change in this PR, but I think it becomes more obvious it's a little unintuitive when it has to get passed through several methods. 

@matt0bi Do you know if there was a specific reason for parsing from event resource name, rather than from the instance data? .As I understand it, they should be the same.

src/gordon_gcp/plugins/enricher.py
```diff
@@ -152,32 +161,49 @@ def __init__(self, config, http_client, success_channel, error_channel):
...
-        instance_data = await self._poll_for_instance_data(
-            event_data['resourceName'], msg_logger)
+    def _get_hostname(self, resource_name, msg_logger):
+        # pulled out for easy subclassing
+        return resource_name.split('/')[-1]
 
-        a_record = await self._create_A_rrecord(event_data, instance_data,
-                                                msg_logger)
-        return [a_record]
+    def _get_fqdn(self, resource_name, msg_logger):
```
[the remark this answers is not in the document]
ocsw: In case a subclass needs it.

src/gordon_gcp/plugins/enricher.py
matt0bi: @descent3me I don't remember any specific reason. Looking at the code, your suggestion is spot on!

src/gordon_gcp/plugins/enricher.py
```diff
@@ -127,19 +127,28 @@ def __init__(self, config, http_client, success_channel, error_channel):
         self._http_client = http_client
         self._logger = logging.getLogger('')
 
+    async def _check_instance_data(self, resource_name, instance_data,
+                                   msg_logger):
```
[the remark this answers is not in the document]
descent3me: Can you give an example? An error is already logged if it fails to load, so  I don't see how a subclass would need to log anything else.  I also think this method should be sync since it should just be checking the response body. No other requests should happen here.

I am arguing against removing so much stuff because I am afraid we are making a more complicated and harder to follow API based on potential use cases, which will make it harder to use and reason about in the use cases that we know exist.

src/gordon_gcp/plugins/enricher.py
```diff
@@ -127,19 +127,21 @@ def __init__(self, config, http_client, success_channel, error_channel):
...
     async def _poll_for_instance_data(self, resource_name, msg_logger):
         exception = None
         backoff = 2
         base_url = f'https://compute.googleapis.com/compute/v1/{resource_name}'
 
-        # Poll until instance data contains external IP information.
+        # Poll until instance data contains all necessary information.
         for attempt in range(1, self.config['retries'] + 1):
             try:
                 msg_logger.debug(f'Attempt {attempt}: fetching {base_url}')
```
descent3me: @ocsw I didn't see this message was already being logged when we were going through the change together. I think this makes your added message redundant.

ocsw:
removed the extra log message

PR #18, opened 2018-05-10
Add config for arbitrary Google zone name

ocsw:
Google has arbitrary zone names in addition to the actual DNS zones.  Add a config so we're not assuming the zone name.

This will have to be propagated to later code, especially the publisher, but I want to surface it.

Fixes #

**Release note**:

```release-note

```


ocsw:
nm, folding this into publisher stuff into another PR
https://github.com/spotify/gordon-gcp/pull/19

PR #19, opened 2018-05-10
Add Publisher plugin for Google Cloud DNS

smstone:
Adds the Publisher plugin for Google Cloud DNS

Fixes # N/A

**Release note**:

```release-note
NONE
```

@spotify/alf

### DIFF
--- a/src/gordon_gcp/exceptions.py
+++ b/src/gordon_gcp/exceptions.py
@@ -30,8 +30,20 @@ class InvalidMessageError(GCPGordonError):
     """Consumed an invalid message from Google Pub/Sub."""
 
 
+class InvalidDNSZoneInMessageError(GCPGordonError):
+    """Raised when a message with an invalid DNS zone is consumed."""
+
+
 class GCPHTTPError(GCPGordonError):
     """An HTTP error occured."""
+
+
+class GCPHTTPConflictError(GCPHTTPError):
+    """An HTTP 409 was received."""
+
+
+class GCPHTTPNotFoundError(GCPHTTPError):
+    """An HTTP 404 was received."""
 
 
 class GCPAuthError(GCPGordonError):
@@ -40,3 +52,7 @@ class GCPAuthError(GCPGordonError):
 
 class GCPConfigError(GCPGordonError):
     """Improper or incomplete configuration for plugin."""
+
+
+class GCPPublishRecordTimeoutError(GCPGordonError):
+    """Time out error when attempting to publish records."""
--- a/src/gordon_gcp/plugins/publisher.py
+++ b/src/gordon_gcp/plugins/publisher.py
@@ -13,23 +13,39 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 """
-Client module to publish DNS records from an event message. Once an
-event message is done (either successfully published, or met with
-errors along the way), it will placed into the appropriate channel,
-either the ``success_channel`` or ``error_channel`` to be further
-handled by the ``gordon`` core system.
+Client module to publish DNS records from an event message. Once an event
+message is done (either successfully published, or met with errors along the
+way), it will placed into the appropriate channel, either the
+`success_channel`` or ``error_channel`` to be further handled by the ``gordon``
+core system.
 
 .. attention::
-    The publisher client is an internal module for the core gordon
-    logic. No other use cases are expected.
+    The publisher client is an internal module for the core gordon logic. No
+    other use cases are expected.
 """
+
+import asyncio
+import datetime
+import json
+import logging
+import re
 
 import zope.interface
 from gordon import interfaces
 
+from gordon_gcp import exceptions
+from gordon_gcp.plugins import _utils
 
 __all__ = ('GDNSPublisher',)
+
+BASE_CHANGES_ENDPOINT = ('https://www.googleapis.com/dns/{version}/projects/'
+                         '{project}/managedZones/{managedZone}/changes')
+RESOURCE_RECORDS_ENDPOINT = ('https://www.googleapis.com/dns/v1/projects/'
+                             '{project}/managedZones/{managedZone}/rrsets')
+# see https://cloud.google.com/dns/api/v1/changes#resource
+DNS_CHANGES_DONE = 'done'
 
 
 @zope.interface.implementer(interfaces.IPublisherClient)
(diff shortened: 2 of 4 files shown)


src/gordon_gcp/clients/http.py
```diff
@@ -114,9 +124,13 @@ def __init__(self, auth_client=None, session=None):
...
         req_kwargs = {
             'params': params,
-            'data': body,
             'headers': req_headers,
         }
+        if body:
+            req_kwargs['data'] = body
+        if json:
+            req_kwargs['json'] = json
+
```
descent3me: If we're keeping this commit, shouldn't we fix the bit on line 139/153?

src/gordon_gcp/plugins/enricher.py
```diff
@@ -79,6 +79,10 @@ def _validate_config(self):
             msg = 'A dns zone must be a FQDN and end with the root zone (".").'
             errors.append(msg)
 
+        if not self.config.get('default_ttl'):
+            msg = 'A default record TTL is required.'
+            errors.append(msg)
+
```
descent3me: I'm not sure this should be in the Publisher PR

src/gordon_gcp/plugins/publisher.py
```diff
@@ -13,44 +13,222 @@
...
+from gordon_gcp import exceptions
+from gordon_gcp.plugins import _utils
 
 __all__ = ('GDNSPublisher',)
 
+HOST = 'https://www.googleapis.com'
+BASE_CHANGES_ENDPOINT = ('{host}/dns/{version}/projects/{project}/managedZones/'
+                         '{managedZone}/changes')
+RESOURCE_RECORDS_ENDPOINT = ('{host}/dns/v1/projects/{project}/managedZones/'
+                             '{managedZone}/rrsets')
```
descent3me: Can we integrate the HOST into the URLs from the start?  It seems like overkill to format it in later since it's the same for all calls.

descent3me: I think we were going to throw version in there too?

src/gordon_gcp/plugins/publisher.py
```diff
@@ -13,44 +13,222 @@
...
+                    raise exceptions.GCPHTTPNotFoundError(e)
+            raise e
+
+        resp_dict = json.loads(resp)
+
+        # TODO: create another task to measure propagation time
+        return resp_dict['id']
+
+    async def _handle_additions_conflict(self, changes):
+        """Compensate for an addition Changes request that returned an HTTP 409.
```
descent3me: I would keep it in the present tense, i.e. `returns`. Also the longer description isn't really a sentence

descent3me: Oh yeah, and I think we were going to add a warning that method calls a URL that returns *ALL* rrsets in a zone, so it could potentially be huge and slow.

src/gordon_gcp/exceptions.py
```diff
@@ -27,16 +27,32 @@ class GCPGordonError(core_exceptions.GordonError):
 
 
 class InvalidMessageError(GCPGordonError):
-    """Consumed an invalid message from Google Pub/Sub."""
+    """Consumed an invalid message from  Pub/Sub."""
```
ocsw: Why remove Google?  And in any case there's now an extra space.

src/gordon_gcp/plugins/enricher.py
ocsw: Yeah, long-term it should be in gordon core.  For now, though...

src/gordon_gcp/plugins/publisher.py
ocsw: ISTR we eventually decided to leave version alone.

src/gordon_gcp/plugins/publisher.py
ocsw: I would also make it 'additions', since that's what Google calls it.

src/gordon_gcp/clients/http.py
```diff
@@ -83,25 +83,35 @@ def __init__(self, auth_client=None, session=None):
...
-            body (obj): (optional) A dictionary, bytes, or file-like
-                object to send in the body of the request.
             headers (dict): (optional) HTTP headers to send with the
                 request. Headers pass through to the request will
                 include :attr:`DEFAULT_REQUEST_HEADERS`.
+            body (obj): (optional) A dictionary, bytes, or file-like
+                object to send in the body of the request.
+            json (obj): (optional) Any json compatible python
+                object.
+                NOTE: json and body parameters could not be used at
```
ocsw: could -> can

src/gordon_gcp/clients/http.py
```diff
@@ -168,3 +182,27 @@ def __init__(self, auth_client=None, session=None):
             json_callback = json.loads
         response = await self.request(method='get', url=url, **kwargs)
         return json_callback(response)
+
+    async def get_all(self, url, params):
+        """Aggregate data from all pages of a API query.
```
ocsw: a -> an

src/gordon_gcp/clients/http.py
```diff
@@ -168,3 +182,27 @@ def __init__(self, auth_client=None, session=None):
...
+        """
+        items = []
+        next_page_token = None
+
+        while True:
+            if next_page_token:
+                params['pageToken'] = next_page_token
+            response = await self.get_json(url, params=params)
+
+            items.append(response)
```
ocsw: Is append here?  I'm not sure about the format, but maybe extend?  Or even something further down the tree?

src/gordon_gcp/plugins/enricher.py
ocsw: Also, I just landed a change that will create a merge conflict here.  You'll want to rebase and fix.

src/gordon_gcp/plugins/publisher.py
```diff
@@ -13,44 +13,222 @@
...
+    def __init__(self, config, success_channel, error_channel, http_client):
         self.config = config
         self.success_channel = success_channel
         self.error_channel = error_channel
+        self.http_client = http_client
+        self.publish_wait_timeout = config['publish_wait_timeout']
+        self.managed_zone = config['managed_zone']
+        self.dns_zone = config['dns_zone']
+        self.project = config['project']
+        self.default_ttl = self.config['default_ttl']
```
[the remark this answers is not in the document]
ocsw: We are also inconsistently handling the http_client.  Enricher does it in the builder...

src/gordon_gcp/plugins/publisher.py
```diff
@@ -13,44 +13,222 @@
...
+
+        resp_dict = json.loads(resp)
+
+        # TODO: create another task to measure propagation time
+        return resp_dict['id']
+
+    async def _handle_additions_conflict(self, changes):
+        """Compensate for an addition Changes request that returned an HTTP 409.
+
+        Get the existing record from the API, append deletion action with the
```
ocsw: I guess 'deletions' here...

tests/unit/clients/test_http.py
```diff
@@ -212,3 +212,68 @@ def simple_json_callback(resp):
...
+
+    post_data = {'hello': 'world'}
+
+    request_kwargs = {
+        'method': 'post',
+        'url': conftest.API_URL,
+        post_arg: post_data
+    }
+
+    with aioresponses() as mocked:
```
ocsw: I probably just need to do some reading, but how does this work?

src/gordon_gcp/plugins/enricher.py
descent3me: I meant in terms of being in enricher, when this PR relates to publisher.

src/gordon_gcp/plugins/publisher.py
descent3me: I thought we had decided to be consistent with version across both URLs?

tests/unit/clients/test_http.py
matt0bi: aioresponses is a nifty library to make it easier to mock http interactions. Instead of having to manually mock and patch things on aiohttp, aioreponses does all that for you. You have to specify the endpoint URL you expect to hit and its response first like the line below. If you were mocking out a get request, it'd look like:

```python
mocked.get('http://example.com', body='<h1>Hello world</h1>')
```

Lemme know if you wanna talk more about it!

Link: https://github.com/pnuckowski/aioresponses

src/gordon_gcp/plugins/publisher.py
```diff
@@ -13,44 +13,222 @@
...
+        try:
+            resp = await self.http_client.request(
+                'post', self.base_changes_url, json=changes)
+        except exceptions.GCPHTTPError as e:
+            error_msg = e.args[0]
+            # Extract status code from error message.
+            match = re.search('\d+', error_msg)
+            if match:
+                status_code = int(match.group(0))
+                if status_code == 409:
```
matt0bi: I dunno if you already talked about this while mobbing, but this is a future-refactoring idea: it seems awkward that the publisher has to parse an error string to figure out why the http client is failing. Since the publisher also has to know a bit about HTTP to know when to drop/retry, maybe the http client could add the status code to the exception?

descent3me: We did talk about it a bit, but we didn't come up with this particular idea, which I think is a good one. Mostly we talked about why the http client does a `raise_for_status` at all, and whether just returning the response (which has the code) would be better (or at least having it configurable).  I think the decision we settled on was that some refactoring was a good idea,  but outside the scope of getting this PR merged.

tests/unit/plugins/test_publisher.py
```diff
@@ -15,22 +15,409 @@
...
+
+
+@pytest.fixture
+def event_msg_data_bad_rrdata(event_msg_data):
+    event_msg_data['resourceRecords'][0]['rrdatas'] = ['127.0.0.899']
+    return event_msg_data
+
+
+@pytest.fixture
+def event_msg_data_updating_record(event_msg_data,
```
ocsw: This seems not to be in use anymore.

tests/unit/plugins/test_publisher.py
```diff
@@ -15,22 +15,409 @@
...
+
+
+@pytest.fixture
+def event_msg_data_with_invalid_zone(event_msg_data):
+    event_msg_data['resourceRecords'][0]['name'] = 'brokenexample.com.'
+    return event_msg_data
+
+
+@pytest.fixture
+def event_msg_data_bad_rrdata(event_msg_data):
```
ocsw: This seems to not be in use any more.

tests/unit/plugins/test_publisher.py
```diff
@@ -15,22 +15,409 @@
...
+
+
+@pytest.fixture
+def publisher_instance(mock_http_client, config):
+    success, error = asyncio.Queue(), asyncio.Queue()
+    pb = publisher.GDNSPublisher(config, success, error, mock_http_client)
+    return pb
+
+
+# TESTS
```
ocsw: Might want to put a box around this to make it more visible in the huge file.

tests/unit/plugins/test_publisher.py
```diff
@@ -15,22 +15,409 @@
...
+    assert 3 == len(caplog.records)
+
+
+@pytest.mark.asyncio
+async def test_publish_changes_raises_exception_on_publish_timeout(
+    publisher_instance, event_message, changes_url,
+        initial_changes, initial_changes_req,
+        initial_changes_pending_json_resp, caplog):
+    """Ensure exception raised when publish wait timeout exceeded."""
+    publisher_instance.publish_wait_timeout = 0.0001
```
ocsw: I'm wondering if this could be flaky...

ocsw:
Still working on the tests a bit, but I'm not sure we have tests for the happy paths in publish_changes and _changes_published?

src/gordon_gcp/plugins/publisher.py
smstone: I'd prefer to leave version as configureable given we might use alpha/beta endpoints.

tests/unit/plugins/test_publisher.py
smstone: Can you elaborate please?

tests/unit/plugins/test_publisher.py
```diff
@@ -15,22 +15,409 @@
...
+            "ttl": 3600
+        }],
+        "startTime": "2018-04-26T15:02:17.541Z",
+        "id": "999",
+        "status": "pending"
+    }"""
+
+
+@pytest.fixture
+def initial_changes():
```
ocsw: I think this should be named initial_changes_done_resp, to match the others.

tests/unit/plugins/test_publisher.py
```diff
@@ -15,22 +15,409 @@
...
+    (all_existing_zone_records_empty(), initial_changes_req())
+]
+
+
+@pytest.mark.parametrize('existing,output', conflict_types)
+@pytest.mark.asyncio
+async def test__handle_additions_conflict(
+        publisher_instance, existing, initial_changes_req, output):
+    """Test correctly handling an additions conflict."""
+    publisher_instance.http_client._get_all_mock.return_value = (existing)
```
ocsw: Why the ()?  I don't think they actually do anything here.  If you wanted to create a tuple, you need a comma.

src/gordon_gcp/plugins/publisher.py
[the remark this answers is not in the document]
ocsw: Ah.  Checked again.  Actually the http_client is passed in the same way in both.  What _is_ different is that there's a builder class for the enricher.  Long-term, we'll have to pull some of this into gordon core and standardize it.

src/gordon_gcp/clients/http.py
```diff
@@ -149,10 +149,14 @@ def __init__(self, auth_client=None, session=None):
 
                     new_req_kwargs = {
                         'params': params,
-                        'body': body,
                         'headers': headers,  # use original req headers
                         'cred_refresh_attempt': refresh_attempt + 1
                     }
+                    if req_kwargs['data']:
+                        new_req_kwargs['data'] = req_kwargs['data']
```
ocsw: That won't work because it's getting passed to request().  It needs to be body.  Honestly, I would just add "'json": json' after 152.  And probably change new_req_args to be something like recursive_args.

descent3me: I think we should scrap the recursive calling altogether. It makes things significantly more complicated for one extra retry (given that MAX_REFRESH_ATTEMPTS=2)

tests/unit/plugins/test_publisher.py
ocsw: I've got it.  Mock datetime.timedelta() and pass -1.

tests/unit/plugins/test_publisher.py
```diff
@@ -15,22 +15,409 @@
...
+        initial_changes_req, handled_conflict_changes_req,
+        initial_changes_pending_json_resp, initial_changes_resp, caplog):
+    """Ensure exception raised on invalid zone"""
+    expected_change_id = json.loads(initial_changes_pending_json_resp)['id']
+    event_message.data['resourceRecords'] = initial_changes_req['additions']
+    publisher_instance.http_client._request_post_mock.side_effect = [
+        exceptions.GCPHTTPError('409'), initial_changes_pending_json_resp]
+    publisher_instance.http_client._get_all_mock.return_value = (
+        all_existing_zone_records)
+    publisher_instance.http_client._get_json_mock.side_effect = [
```
ocsw: This works, but could be a return_value.

tests/unit/plugins/test_publisher.py
```diff
@@ -15,22 +15,409 @@
...
+
+@pytest.mark.asyncio
+async def test_publish_changes_raises_exception_on_publish_timeout(
+    publisher_instance, event_message, changes_url,
+        initial_changes_req, initial_changes_resp,
+        initial_changes_pending_json_resp, mock_sleep, caplog):
+    """Ensure exception raised when publish wait timeout exceeded."""
+    publisher_instance.publish_wait_timeout = 0.0001
+    event_message.data['resourceRecords'] = initial_changes_req['additions']
+    publisher_instance.http_client._request_post_mock.side_effect = [
```
ocsw: same re return_value

tests/unit/plugins/test_publisher.py
```diff
@@ -15,22 +15,409 @@
...
+        initial_changes_pending_json_resp]
+    initial_changes_resp['status'] = 'pending'
+    publisher_instance.http_client._get_json_mock.side_effect = [
```
ocsw: same re return_value

tests/unit/plugins/test_publisher.py
```diff
@@ -15,22 +15,409 @@
...
+    assert 3 == len(caplog.records)
+
+
+@pytest.mark.asyncio
+async def test_publish_changes_raises_exception_on_publish_timeout(
+    publisher_instance, event_message, changes_url,
+        initial_changes_req, initial_changes_resp,
+        initial_changes_pending_json_resp, mock_sleep, caplog):
+    """Ensure exception raised when publish wait timeout exceeded."""
+    publisher_instance.publish_wait_timeout = 0.0001
```
ocsw: Did we want to try the mocking datetime approach?

tests/unit/plugins/test_publisher.py
smstone: ah good point - i think i originally had a few side effects

tests/unit/plugins/test_publisher.py
smstone: i'm OK with leaving this as is.

PR #20, opened 2018-05-15
Enricher builder hook

ocsw:
Factor out validations for easier subclassing.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/src/gordon_gcp/plugins/enricher.py
+++ b/src/gordon_gcp/plugins/enricher.py
@@ -62,32 +62,50 @@ class GCEEnricherBuilder:
         self.success_channel = success_channel
         self.error_channel = error_channel
         self.kwargs = kwargs
+        self._validate_config()
+        self.http_client = self._init_http_client()
+
+    def _validate_keyfile(self):
+        msg = []
+        if not self.config.get('keyfile'):
+            msg.append('The path to a Service Account JSON keyfile is required '
+                       'to authenticate to the GCE API.')
+        return msg
+
+    def _validate_dns_zone(self):
+        msg = []
+        if not self.config.get('dns_zone'):
+            msg.append('A dns zone is required to build correct A records.')
+        if not self.config.get('dns_zone', '').endswith('.'):
+            msg.append('A dns zone must be an FQDN and end with the root '
+                       'zone (".").')
+        return msg
+
+    def _validate_retries(self):
+        if not self.config.get('retries'):
+            self.config['retries'] = 5
+        return []
+
+    def _call_validators(self):
+        """Actually run all the validations.
+
+        Returns:
+            list(str): Error messages from the validators.
+        """
+        msg = []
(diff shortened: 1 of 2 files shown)


src/gordon_gcp/plugins/enricher.py
```diff
@@ -57,37 +58,52 @@ class GCEEnricherBuilder:
         kwargs (dict): Additional keyword arguments to pass to the
             enricher.
     """
+    VALIDATORS = [
+        gordon_gcp.enricher.GCEEnricherBuilder._validate_keyfile,
+        gordon_gcp.enricher.GCEEnricherBuilder._validate_dns_zone,
+        gordon_gcp.enricher.GCEEnricherBuilder._validate_retries
```
smstone: Do these need to be fully qualified here?

src/gordon_gcp/plugins/enricher.py
```diff
@@ -57,37 +59,52 @@ class GCEEnricherBuilder:
...
+        msg = []
+        if not config.get('dns_zone'):
+            msg.append = 'A dns zone is required to build correct A records.'
+        if not config.get('dns_zone', '').endswith('.'):
+            msg.append = ('A dns zone must be a FQDN and end with the root '
+                          'zone (".").')
+        return msg
+
+    @staticmethod
+    def _validate_retries(config):
```
descent3me: Could these methods be abstracted into something like `_validate_config(key, err_msg, default=None)` that handles the logging and error raising? That could then be abstracted out since it would be useful to other plugins. I think it's fine to error out on the first missing key, if that is the side effect.

src/gordon_gcp/plugins/enricher.py
```diff
@@ -32,8 +32,10 @@
 import logging
 
 import zope.interface
+
 from gordon import interfaces
 
+import gordon_gcp
```
descent3me: We should be more specific here.

src/gordon_gcp/plugins/enricher.py
ocsw: They do.  You can't have a class variable refer to static methods without the class name, and that isn't defined yet during the class definition.

src/gordon_gcp/plugins/enricher.py
ocsw: Hmmmmm.  That's not a bad idea.  It also solves the pathing problem in VALIDATORS.

descent3me: I'd personally be fine with skipping it. If you'd like to keep it, consider it an edge case and just do the check after calling the original validate method. It requires a little code duplication, but still less than it is now.

ocsw:
Addressed comments.  Ok?

descent3me:
I would rename the commit, but otherwise looks ok from my side

PR #21, opened 2018-05-16
Fix http client token refreshing logic

matt0bi:
This PR fixes a bug where the http client refreshes the auth token on every request sent to a Google API because `self._auth_client.creds.token` is _always_ None. This is related to this old fix: https://github.com/spotify/gordon-janitor-gcp/pull/10/files

N/A

N/A

**Release note**:

```release-note
NONE
```

@spotify/alf

### DIFF
--- a/src/gordon_gcp/clients/auth.py
+++ b/src/gordon_gcp/clients/auth.py
@@ -41,10 +41,10 @@ To use:
     >>> auth_client = google_gcp.GAuthClient(keyfile=keyfile)
     # with Application Default Credentials
     >>> auth_client = google_gcp.GAuthClient()
-    >>> auth_clientcreds.token is None
+    >>> auth_client.token is None
     True
     >>> loop.run_until_complete(auth_client.refresh_token())
-    >>> auth_client.creds.token
+    >>> auth_client.token
     'c0ffe3'
 
 """
--- a/src/gordon_gcp/clients/http.py
+++ b/src/gordon_gcp/clients/http.py
@@ -73,11 +73,11 @@ class AIOConnection:
         """Check for validity of token, and refresh if none or expired."""
         is_valid = False
 
-        if self._auth_client.creds.token:
+        if self._auth_client.token:
             # Account for a token near expiration
             now = datetime.datetime.utcnow()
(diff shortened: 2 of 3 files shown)


descent3me:
+1 to the bug fix, but i think the test needs more work.  this kind of a bug should have caused a test to fail.

descent3me:
Also there's some lines here that should be updated
https://github.com/spotify/gordon-gcp/blob/develop/src/gordon_gcp/clients/auth.py#L44

smstone:
@matt0bi / @descent3me I refactored the http module via https://github.com/spotify/gordon-gcp/pull/19 I also added the changes you have here (minus the tests, which I believe I covered). Can you PTAL?

descent3me:
The test you added is better than the one that was there before, because it captures both sides of the if-condition. However, both it and the previously existing test are trying to test *too much*.  It is testing the internals of the auth client, which means you are mocking them out, which is why the previous test didn’t raise an error.  We shouldn’t be messing with the internal member variables of the GAuthClient, or worse yet with those variables internal member variables (as was the case with setting `auth_client.creds.token`), in a unit test for the AIOConnection.  Because we tried to mock those out, we inadvertently added those variables to auth_client, so an AttributeError wasn’t raised. 

Probably part of the reason this happened is because the separation of logic between AIOConnection and GAuthClient isn’t clean.  I think the best fix would be to move the entire `set_valid_token` method to the GAuthClient, because the http client shouldn’t have to care about tokens or expiry.  There should be only one public method like `auth_client.authorize()` or `auth_client.assert_authorized()` (or whatever) and then the auth client itself decides whether a new token needs to be created.  In addition to a better separation of logic, when the method is tested it will be along with the other auth client unit tests, which should be using a real object, and it's much less likely we'll accidentally add attributes.

matt0bi:
@descent3me @smstone I think @smstone's PR is more comprehensive and a better step towards eventually refactoring http/auth, so I'm going to close this PR and focus on #22. 

@descent3me Thanks for the writeup, it's really good. I think we should do that refactoring when we're working on v2 (or a hackday?), so I added a note and a link to your comment to the gordon todo doc.

PR #23, opened 2018-05-17
fix config to properly show dns zone vars

ocsw:
The code uses dns_zone in multiple places, and also has managed_zone.  They should be in the sample config in the correct places.

Fixes #

**Release note**:

```release-note

```


matt0bi:
:+1:

PR #24, opened 2018-05-30
Merge gordon-janitor-gcp repo into gordon-gcp

ocsw:
We want to consolidate the GCP plugins so we have fewer repos and we avoid having code duplicated or in the wrong place.

Fixes #

**Release note**:

```release-note
Merged gordon-janitor-gcp repo into gordon-gcp
```


CHANGELOG.rst
```diff
@@ -1,8 +1,10 @@
 Changelog
 =========
 
-0.0.1.dev3 (2018-05-22)
------------------------
+0.0.1.dev3 (2018-05-22 / 2018-05-25)
```
econchick: Two things: 1. I think this should be edited once a release is made (aka not in this PR/branch), and 2. Once we do make a release, I think we shouldn't merge the two histories like this. Rather, just have a line item under the released version saying that the repos have been merged. We can point to the change log of the janitor GCP if folks are curious.

docs/plugins.rst
```diff
@@ -26,12 +26,35 @@ Event Consumer
 .. autoclass:: gordon_gcp.GPSEventConsumer
     :members:
 
-Publisher
-----------
+GDNS Publisher
```
econchick: I think it'd make more sense to divide the docs into sections for the service vs janitor. 

Relatedly, I think we should separate the plugins w/i the repo itself, e.g. `gordon_gcp.plugins.service.<plugin>` and `gordon_gcp.plugins.janitor.<plugin>`. It will avoid the name spacing issues that you encountered, and be a clearer separation (like we discussed yesterday afternoon).

CHANGELOG.rst
econchick: Also FWIW we can't re-release a version since PyPI does not allow that. We are going to have to bump the release.

docs/plugins.rst
ocsw: Yeah I had originally forgotten to do that.  The question is, if I do that, should I also do it for clients?  Some are specific to the janitor.  I'd prefer not, but...

CHANGELOG.rst
ocsw: Yeah I know this became a mess.  I think you're right.  I'll also add a commit to bump the version.

docs/plugins.rst
ocsw: Do you mean divide each doc, or have separate docs?

ocsw: Oh, and regarding namespacing, I kinda think we should rename them anyway for clarity.  Too easy to look for a file and get the wrong one.

src/gordon_gcp/plugins/__init__.py
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright 2017 Spotify AB
+# Copyright 2018 Spotify AB
```
econchick: nitpick, but from a FOSS board perspective, I prefer a date range. If it weren't a date range, it should be 2017.

ocsw: Even if the file in question in from 2018?  This was actually a thing I wanted to ask about.  I will say that if we use ranges, people will 100% never update them.

econchick: The former, the git diff makes it look like a file that's changed, not a new file. Re: the latter, I'm fine if people forget, that's why I said if it weren't a date range, then 2017.

docs/plugins.rst
ocsw: We discussed on Slack.  Results: I divided the plugins doc, but the rest are ok.  I split the directories for (only) the plugins.  We're leaving the new namespacing in place.

src/gordon_gcp/plugins/__init__.py
ocsw: Discussion results: if a file is entirely new, use the current year.  In this case, there was a file there, but it was moved to service/ and this file was created from scratch.

ocsw: Further results: I'll change the dates.

PR #26, opened 2018-06-01
Add builder class + entrypoint func for GDNSPublisher

econchick:

As it stands, core will pass in only the config and channels to each plugin. But the `GDNSPublisher` is expecting an HTTP client as well. It also does not gracefully handle required config keys like all other plugins (and cause unhandled `KeyErrors`).

This PR adds a builder class to validate config, create the auth+http clients, and instantiates a `GDNSPublisher`. In keeping with the API design/abstraction layers, this also adds the helper function for the `setup.py` entrypoint.

I've also updated the docs on configuration to reflect required config keys (for all plugins).

N/A

**Release note**:

```release-note
Fix failure for core to instantiate GDNSPublisher plugin.
```

@spotify/alf

### DIFF
--- a/docs/config-janitor.rst
+++ b/docs/config-janitor.rst
@@ -1,5 +1,5 @@
-Configuration
-=============
+Janitor Configuration
+=====================
 
 Configuring Google Cloud Platform :doc:`plugins` for the `gordon-janitor`_ service.
 
--- a/docs/config-service.rst
+++ b/docs/config-service.rst
@@ -1,5 +1,5 @@
-Configuration
-=============
+Service Configuration
+=====================
 
 Configuring Google Cloud Platform :doc:`plugins` for the `gordon`_ service.
 
--- a/docs/config-service.rst
+++ b/docs/config-service.rst
@@ -48,22 +48,64 @@ Plugin Configuration
 ``[gcp.event_consumer]``
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-All configuration options above in the general ``[gcp]`` may be used here. There are no specific Google Pub/Sub Consumer-related configuration options.
+All configuration options above in the general ``[gcp]`` may be used here. Additional Google Pub/Sub Consumer-related configuration options are:
 
+.. option:: topic="STR"
+
+    `Required`: A topic to which the Event Consumer client must subscribe.
+
+    For more information on Google Pub/Sub topics, please see `Google's docs on managing topics <topics>`_.
+
+.. option:: subscription="STR"
+
+    `Required`: A subscription to the ``topic`` from which the Event Consumer client will pull.
+
+    For more information on Google Pub/Sub subscriptions, please see `Google's docs on managing subscriptions <subscriptions>`_.
 
 ``[gcp.enricher]``
 ~~~~~~~~~~~~~~~~~~
 
-All configuration options above in the general ``[gcp]`` may be used here. There are no specific Google Compute Engine Enricher-related configuration options.
+All configuration options above in the general ``[gcp]`` may be used here. Additional Google Compute Engine configuration options are:
+
+
+.. option:: dns_zone="STR"
+
+    `Required`: DNS zone to validate the correctness of A records before publishing. Must be a fully-qualified domain name (FQDN), ending in ``.``, e.g. ``example.com.``.
 
 
 ``[gcp.publisher]``
 ~~~~~~~~~~~~~~~~~~~
 
-All configuration options above in the general ``[gcp]`` may be used here. There are no specific Google DNS Publisher-related configuration options.
+All configuration options above in the general ``[gcp]`` may be used here. Additional Google Cloud DNS configuration options are:
 
+.. describe:: dns_zone="STR"
+
+    `Required`: DNS zone to validate the correctness of A records before publishing. Must be a fully-qualified domain name (FQDN), ending in ``.``, e.g. ``example.com.``.
+
+.. option:: managed_zone="STR"
+
+    `Required`: The managed zone name in Google Cloud DNS where records are to be published.
+
+    To learn more about managed zones, please see `Google's docs on managed zones <managed_zones>`_.
+
+.. option:: default_ttl=INT
+
+    `Required`: The default TTL in seconds. This will be used if the publisher receives a record set to be published that does not yet have the TTL set. Must be greater than 4.
+
+.. option:: publish_wait_timeout=INT|FLOAT
+
+    `Optional`: Timeout in seconds for waiting for confirmation that changes have been successfully completed within Google Cloud DNS. Default is 60 seconds.
+
+.. option:: api_version="STR"
+
+    `Optional`: API version for both the `changes`_ endpoint and the `resource records`_ endpoint.
 
 
 .. _`gordon`: https://github.com/spotify/gordon
 .. _`keyfiles`: https://cloud.google.com/iam/docs/creating-managing-service-account-keys
 .. _`projects`: https://cloud.google.com/resource-manager/docs/creating-managing-projects
+.. _`topics`: https://cloud.google.com/pubsub/docs/admin#managing_topics
+.. _`subscriptions`: https://cloud.google.com/pubsub/docs/admin#managing_subscriptions
+.. _`managed_zones`: https://cloud.google.com/dns/zones/
+.. _`changes`: https://cloud.google.com/dns/api/v1/changes
+.. _`resource records`: https://cloud.google.com/dns/api/v1/resourceRecordSets/list
--- a/src/gordon_gcp/plugins/service/__init__.py
+++ b/src/gordon_gcp/plugins/service/__init__.py
@@ -16,17 +16,18 @@
 
 from gordon_gcp.plugins.service import enricher
 from gordon_gcp.plugins.service import event_consumer
+from gordon_gcp.plugins.service import gdns_publisher
 # Mainly for easier documentation reading
-from gordon_gcp.plugins.service.enricher import *  # noqa: F403
-from gordon_gcp.plugins.service.event_consumer import *  # noqa: F403
-from gordon_gcp.plugins.service.gdns_publisher import *  # noqa: F403
+from gordon_gcp.plugins.service.enricher import *  # noqa: F401,F403
+from gordon_gcp.plugins.service.event_consumer import *  # noqa: F401,F403
+from gordon_gcp.plugins.service.gdns_publisher import *  # noqa: F401,F403
 
 
 __all__ = (
     enricher.__all__ +  # noqa: F405
     event_consumer.__all__ +  # noqa: F405
     gdns_publisher.__all__ +  # noqa: F405
-    ('get_event_consumer', 'get_enricher')
+    ('get_event_consumer', 'get_enricher', 'get_gdns_publisher')
 )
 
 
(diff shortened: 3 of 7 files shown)


docs/config-janitor.rst
```diff
@@ -1,5 +1,5 @@
-Configuration
-=============
+Janitor Configuration
+=====================
```
econchick: Without this differentiation, the ToC would be this:

2018-06-01 at 4 08 pm

docs/config-service.rst
```diff
@@ -48,22 +48,64 @@ Plugin Configuration
...
+    `Required`: DNS zone to validate the correctness of A records before publishing. Must be a fully-qualified domain name (FQDN), ending in ``.``, e.g. ``example.com.``.
 
 
 ``[gcp.publisher]``
 ~~~~~~~~~~~~~~~~~~~
 
-All configuration options above in the general ``[gcp]`` may be used here. There are no specific Google DNS Publisher-related configuration options.
+All configuration options above in the general ``[gcp]`` may be used here. Additional Google Cloud DNS configuration options are:
+
+.. describe:: dns_zone="STR"
```
econchick: Originally I had this as `.. option::` but Sphinx warns/fails because it's also defined under the `[gcp.enricher]` section. I thought about putting it under `[gcp]` but that may be a bit misleading since the event consumer doesn't use/need it. `.. describe::` looks the same when rendering so ¯\\_(ツ)_/¯ .

src/gordon_gcp/plugins/service/__init__.py
```diff
@@ -16,17 +16,18 @@
 
 from gordon_gcp.plugins.service import enricher
 from gordon_gcp.plugins.service import event_consumer
+from gordon_gcp.plugins.service import gdns_publisher
 # Mainly for easier documentation reading
-from gordon_gcp.plugins.service.enricher import *  # noqa: F403
-from gordon_gcp.plugins.service.event_consumer import *  # noqa: F403
-from gordon_gcp.plugins.service.gdns_publisher import *  # noqa: F403
+from gordon_gcp.plugins.service.enricher import *  # noqa: F401,F403
```
econchick: I have no idea why once I added L19 that these started to fail (unused import blahblahblah). freakin' flake8 ಠ_ಠ.

tests/unit/plugins/service/test_init.py
```diff
@@ -252,3 +252,98 @@ def test_get_enricher_config_bad_dns_zone(mocker, caplog, enricher_config,
...
+    mocker.patch(patch)
+
+    publisher_config.pop(conf_key)
+    success_chnl, error_chnl = asyncio.Queue(), asyncio.Queue()
+
+    with pytest.raises(exceptions.GCPConfigError) as e:
+        service.get_gdns_publisher(publisher_config, success_chnl, error_chnl)
+
+    e.match('Invalid configuration:\n' + exp_msg_snip)
+    assert 1 == len(caplog.records)
```
ocsw: Should this also test for >1 error?

src/gordon_gcp/plugins/service/gdns_publisher.py
```diff
@@ -48,6 +50,101 @@
 DNS_CHANGES_DONE = 'done'
 
 
+class GDNSPublisherBuilder:
+    """Build and configure a :class:`GDNSPublisher` object.
+
+    Args:
+        config (dict): Google Cloud DNS API related configuration.
+        success_channel (asyncio.Queue): queue to place a successfully
```
ocsw: capital Q

src/gordon_gcp/plugins/service/gdns_publisher.py
```diff
@@ -48,6 +50,101 @@
...
+            published message to be further handled by the ``gordon``
+            core system.
+        error_channel (asyncio.Queue): queue to place a message met
```
ocsw: capital Q

src/gordon_gcp/plugins/service/gdns_publisher.py
```diff
@@ -48,6 +50,101 @@
...
+        kwargs (dict): Additional keyword arguments to pass to the
+            publisher.
+    """
+    def __init__(self, config, success_channel, error_channel, **kwargs):
+        self.config = config
+        self.success_channel = success_channel
+        self.error_channel = error_channel
+        self.kwargs = kwargs
+
+    def _validate_config(self):
```
ocsw: I'd like to do what I did in the enricher so it's easier to subclass (individual test methods and one caller method).

docs/config-service.rst
```diff
@@ -48,22 +48,64 @@ Plugin Configuration
...
+
+.. option:: managed_zone="STR"
+
+    `Required`: The managed zone name in Google Cloud DNS where records are to be published.
+
+    To learn more about managed zones, please see `Google's docs on managed zones <managed_zones>`_.
+
+.. option:: default_ttl=INT
+
+    `Required`: The default TTL in seconds. This will be used if the publisher receives a record set to be published that does not yet have the TTL set. Must be greater than 4.
```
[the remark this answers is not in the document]
econchick: I did some testing via the console, you can put in integers 0-4, but when `dig`'ing, it returns 5. (relates to [this comment](https://github.com/spotify/gordon-gcp/pull/26/commits/a6d767c9a550753b4d701b432ab3103e813d465c#diff-8f53676280c6a88b173ebd83ba0803c0R120)

docs/config-service.rst
```diff
@@ -48,22 +48,64 @@ Plugin Configuration
...
 ~~~~~~~~~~~~~~~~~~~
 
-All configuration options above in the general ``[gcp]`` may be used here. There are no specific Google DNS Publisher-related configuration options.
+All configuration options above in the general ``[gcp]`` may be used here. Additional Google Cloud DNS configuration options are:
+
+.. describe:: dns_zone="STR"
+
+    `Required`: DNS zone to validate the correctness of A records before publishing. Must be a fully-qualified domain name (FQDN), ending in ``.``, e.g. ``example.com.``.
+
+.. option:: managed_zone="STR"
```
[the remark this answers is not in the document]
econchick: But it's only relevant to the publisher; it may not be used if one implements a different publisher that is in GCP (publish to... Datastore? ¯\\_(ツ)_/¯)

src/gordon_gcp/plugins/service/gdns_publisher.py
```diff
@@ -69,12 +166,12 @@ def __init__(self, config, success_channel, error_channel, http_client):
...
         self.error_channel = error_channel
         self.http_client = http_client
-        self.publish_wait_timeout = self.config['publish_wait_timeout']
+        self.publish_wait_timeout = self.config.get('publish_wait_timeout', 60)
         self.managed_zone = self.config['managed_zone']
         self.dns_zone = self.config['dns_zone']
         self.project = self.config['project']
         self.default_ttl = self.config['default_ttl']
-        self.api_version = self.config.get('version', 'v1')
+        self.api_version = self.config.get('api_version', 'v1')
```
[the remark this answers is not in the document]
econchick: Do you mean [these docs](https://github.com/spotify/gordon-gcp/pull/26/commits/1b6a392e88d5fd07d2241dd6be24d9d7e41d2191#diff-6402775da9e40b9d0c6dd4a914669ed6R99)?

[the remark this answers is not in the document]
ocsw: Derp.  Yes, those. :)

ocsw:
Re: config locations, fair enough.  I guess put them in the plugin sections like you have here and update the sample file.  Maybe add a note in the sample file to keep them in sync?

src/gordon_gcp/plugins/service/gdns_publisher.py
```diff
@@ -48,6 +50,101 @@
...
+                   '(".").')
+            errors.append(msg)
+
+        if not self.config.get('managed_zone'):
+            msg = ('A managed zone is required to publish records to Google '
+                   'Cloud DNS.')
+            errors.append(msg)
+
+        try:
+            timeout = float(self.config.get('publish_wait_timeout', 60))
```
descent3me: It seems potentially confusing to have a default here.  If you want to do an float like this, maybe use something more obvious it doesn't matter what it is.  0 maybe?

econchick: Ah, confusing because the default is essentially thrown away? I feel like it might be more confusing if it's 0 here, and [60 here](https://github.com/spotify/gordon-gcp/pull/26/files/5b70925ab427e9695dfc66d82477420e8314b06d#diff-8f53676280c6a88b173ebd83ba0803c0R169). 

I did `float` because timedelta does accept it, and I figured for the potential to want to manage timeout a little more fine-grained. But I'm just as content w/ forcing integers.

econchick:
@ocsw I think I got everything. Making sure the config validation is easier to subclass is a good idea, though I did the config validation a little differently than the enricher. With this approach, you can just append, remove, or overwrite a list of functions that will be called. 

@descent3me LMK what you think about the int v float.

ocsw:
I originally had a list like that and @descent3me shot it down.  Although yours is probably better than mine was.  @descent3me thoughts?  And how much do we care about consistency?

econchick:
I'm +1 for consistency - let's decide on what we want, and then switch the other two in the future.

descent3me:
I think its slightly better since its at least an initialized list, rather than a class constant (which iirc yours was @zephyr?). I believe my suggestion was to use the single validate_config function to call all the specific validate functions directly, rather than having a variable hold them. It makes the builder slightly simpler, since that var is not used anywhere outside that function anyways, which I think could apply here as well @econchick.  

Unrelated to this specific PR, it seems like we have a lot of code for config validation. My opinion is that unless there's some specific thing we're checking (trailing '.', number in a certain range, etc.), we should just default to a KeyError thrown by the init of the plugin itself, logged or outputted in a way the user knows what class threw it, so they know which config to change (probably in the plugin loader, rather than a builder since it is repeated code across all builders). If we need a longer description of why a config parameter is needed, I think that should live in docs to keep the code simpler.

Since there's been some discussion about standardizing how we do config validation, but there's been no consensus that I know of, I don't think we're ready for consistency. I assumed that was something we were going to handle later with factoring out of all the plugins in one go.

descent3me:
@econchick Just got your comment. If you think we're ready for a decision, I'm fine with taking it now, I just didn't want to block your relatively unrelated PR on it.

ocsw:
I'm fine with putting consistifying off for now if that's what we want to do.

My only remaining thing is adding a test for if 2 validations fail at once.

src/gordon_gcp/plugins/service/gdns_publisher.py
```diff
@@ -48,6 +50,116 @@
...
+
+    def _validate_managed_zone(self, errors):
+        if not self.config.get('managed_zone'):
+            msg = ('A managed zone is required to publish records to Google '
+                   'Cloud DNS.')
+            errors.append(msg)
+
+    def _validate_publish_timeout(self, errors):
+        try:
+            timeout = float(self.config.get('publish_wait_timeout', 60))
```
descent3me: My comment wasn't so much about float v int, but that instead casting and type errors to infer numericalness, we could do something like
```python
if 'publish_wait_timeout' in self.config:
    timeout = self.config.get('publish_wait_timeout')
    if not isinstance(timeout, numbers.Number) or timeout < 0:
        msg =  '"publish_wait_timeout" must be a positive number."
        errors.append(msg)
```
It would have the added benefit of not supporting number-like strings, like `publish_wait_timeout = "3"`, which I think is a good thing to avoid.

econchick:
@ocsw I did update the test to test for more than one errors.

ocsw:
@econchick Ah, I missed it because I was looking at the wrong function.  I was thinking of your +1 to my 'Should this also test for >1 error?' in test_get_gdns_publisher_bad_config(), but I guess we don't need to test more than one of those at once if we have it in another function.

ocsw:
Ok, I'm good if @descent3me is.

econchick:
Just to be super explicit so we're on the same page, @ocsw [this was what I was referring to](https://github.com/spotify/gordon-gcp/blob/feature/add-dnspub-builder/tests/unit/plugins/service/test_init.py#L308-L311) w/r/t updating that test you commented on for >1 error.

econchick:
@descent3me pushed an update, mind a final look? @ocsw if it looks okay to you, mind updating the review?

descent3me:
lgtm +1

econchick:
🙏 🌮 \o/

PR #28, opened 2018-06-07
Refactor record marshaling and processing in GDNS publisher

ocsw:
We need to separate the marshaling of resource records and the processing of each record so we can change the marshaling in subclasses.

Fixes #

**Release note**:

```release-note
Refactored GDNSPublisher for easy subclassing/overriding.
```

### DIFF
--- a/src/gordon_gcp/plugins/service/gdns_publisher.py
+++ b/src/gordon_gcp/plugins/service/gdns_publisher.py
@@ -227,6 +227,7 @@ class GDNSPublisher:
 
         Raises:
             GCPHTTPConflictError on HTTP 409.
+            GCPHTTPNotFoundError on HTTP 404.
         """
         try:
             resp = await self.http_client.request(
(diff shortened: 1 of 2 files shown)


ocsw:
@spotify/alf ok?  This is a prereq for https://ghe.spotify.net/gordon/gordon-internal/pull/9.

src/gordon_gcp/plugins/service/gdns_publisher.py
```diff
@@ -298,6 +299,38 @@ def _format_resource_record_changes(self, action, resource_record):
 
         return False
 
+    async def _process_resource_record(self, resource_record, action, logger):
+        """Publish changes for one recordd, making sure they are completed.
```
descent3me: typo

ocsw:
@descent3me You wanted me to change the name of the factored out function.  How's this?

src/gordon_gcp/plugins/service/gdns_publisher.py
```diff
@@ -298,6 +299,39 @@ def _format_resource_record_changes(self, action, resource_record):
 
         return False
 
+    async def _dispatch_changes(
+            self, resource_record, action, logger):
```
descent3me: I know its defined in the spec, but I'd prefer is we could do the line breaking at least to a de-facto pattern. I think we tend to start listing the args on the same line as the method declaration.

matt0bi: ... mumble [yapf](https://github.com/google/yapf) mumble mumble ... :)

ocsw: Ha, it actually fits on one line now.

descent3me:
+1

PR #30, opened 2018-06-08
Change some variable handling in the GDNS publisher

ocsw:
Changes where/if certain variables are set to make it easier to subclass and override.

Fixes #

**Release note**:

```release-note
Internal API improvements.
```

### DIFF
--- a/src/gordon_gcp/plugins/service/gdns_publisher.py
+++ b/src/gordon_gcp/plugins/service/gdns_publisher.py
@@ -41,14 +41,8 @@ from gordon_gcp.clients import auth
 from gordon_gcp.clients import http
 from gordon_gcp.plugins import _utils
 
-__all__ = ('GDNSPublisher', 'GDNSPublisherBuilder')
 
-BASE_CHANGES_ENDPOINT = ('https://www.googleapis.com/dns/{version}/projects/'
-                         '{project}/managedZones/{managedZone}/changes')
-RESOURCE_RECORDS_ENDPOINT = ('https://www.googleapis.com/dns/v1/projects/'
-                             '{project}/managedZones/{managedZone}/rrsets')
-# see https://cloud.google.com/dns/api/v1/changes#resource
-DNS_CHANGES_DONE = 'done'
+__all__ = ('GDNSPublisher', 'GDNSPublisherBuilder')
 
 
 class GDNSPublisherBuilder:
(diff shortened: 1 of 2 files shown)


ocsw:
@spotify/alf ok?

src/gordon_gcp/plugins/service/gdns_publisher.py
```diff
@@ -275,18 +271,19 @@ def _format_resource_record_changes(self, action, resource_record):
...
         See DNS_CHANGES_DONE.
 
         Args:
             change_id (str): Change ID of the Changes resource.
+            base_changes_url (str): The URL prefix to use to check on changes.
 
         Returns:
             True if the changes are done, False if anything else.
         """
-        change_id_url = f'{self.base_changes_url}/{change_id}'
```
descent3me: Why not pass the zone name instead of urls? Then you can just do the formatting once, rather than twice.

ocsw: You mean do it in this function?  The base_changes_url is used twice, so I didn't want to duplicate the code.  I suppose the resource_records_url is only used once, so it might make sense there.

src/gordon_gcp/plugins/service/gdns_publisher.py
```diff
@@ -179,16 +179,9 @@ def __init__(self, config, success_channel, error_channel, http_client):
...
-        self.managed_zone = self.config['managed_zone']
-        self.dns_zone = self.config['dns_zone']
         self.project = self.config['project']
         self.default_ttl = self.config['default_ttl']
         self.api_version = self.config.get('api_version', 'v1')
-        self.base_changes_url = BASE_CHANGES_ENDPOINT.format(
-            version=self.api_version, project=self.project,
-            managedZone=self.managed_zone)
-        self.resource_records_url = RESOURCE_RECORDS_ENDPOINT.format(
-            project=self.project, managedZone=self.managed_zone)
```
descent3me: Maybe these constants should `GDNSPublisher` class constants rather than module constants, so they can be inherited as well.

ocsw: Good idea.

src/gordon_gcp/plugins/service/gdns_publisher.py
descent3me: We talked about this outside of the PR and decided doing a minimal change (i.e. just the url) was better than a larger refactor for now.

descent3me:
+1

PR #32, opened 2018-06-12
Remove routing logic from plugins

descent3me:
To be consistent with the new routing logic which handles phase updates and error handling, plugins should avoid routing directly.

Fixes #

@spotify/alf 

**Release note**:

```release-note
Fix: remove routing logic from plugins
```

### DIFF
--- a/src/gordon_gcp/plugins/service/enricher.py
+++ b/src/gordon_gcp/plugins/service/enricher.py
@@ -140,8 +140,6 @@ class GCEEnricher:
 
     def __init__(self, config, http_client, success_channel, error_channel):
         self.config = config
-        self.success_channel = success_channel
-        self.error_channel = error_channel
         self._http_client = http_client
         self._logger = logging.getLogger('')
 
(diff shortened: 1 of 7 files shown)


econchick:
I'm in the process of reviewing this, but in the future, would you mind terribly in sticking to the gitflow branch naming conventions? Unless it's hotfixes (i.e. fixes that go into the release branch), fixes like this (i.e. not a production fix that needs to be done right away) would still be a `feature` (gitflow author's explanation within [his post](https://nvie.com/posts/a-successful-git-branching-model/)). I guess a better name would be `topic` instead of `feature` but ¯\\_(ツ)_/¯   No need to close/reopen this one though.

descent3me:
@econchick Can you specify the part that goes over the naming you mean?  I read through the contributing guide and https://nvie.com/posts/a-successful-git-branching-model/#supporting-branches before I created the PR to be sure I renamed it correctly (I initially had it as `fix/raise-errors`).  
> Branch naming convention:
anything except master, develop, release-*, or hotfix-*

I was unsure if I should consider it a feature at all, but decided it definitely wasn't a hotfix.

econchick:
@descent3me I can see how it's confusing. How I interpreted it was that it's referring to the prefix (`hotfix/<name>`, `release/<version>`, `feature/<name>`). The author of the git flow idea/that linked post made the [`gitflow`](https://github.com/nvie/gitflow) git tool, where it refers to [another blogpost](https://jeffkreeftmeijer.com/git-flow/) on how to use git flow. Both that post and the [gitflow init setup](https://github.com/nvie/gitflow/blob/develop/git-flow-init#L246) shows that it's the prefix name that doesn't change, that it's an agreed-upon prefix for naming everything that's not a hotfix or a release branch. I agree that it does cause initial confusion; I admit I didn't think of non-hotfix patches and just went with the default (I initially started to use that `gitflow` tool myself when starting these repos, but stopped because we don't use the model internally). And we certainly can be a bit clearer in our contributing docs.

descent3me:
I don't mind adding a line to a contrib doc for future ref, because I couldn't figure it out from what's there now. It looks like right now we've been doing a mix (`feature/blah`, `<feature-description>`, `release-0.0.1.dev3`) - is it prefix + '/'  you want to go with (i.e. your examples)?

PR #37, opened 2018-06-19
Bug fixes and improvements from trial running plugins with gordon service

econchick:

In trying to get the service running with @smstone, we encountered various bugs and things that could be improved.

I tried to make the commit subjects + messages clear, so I won't copy-pasta here.

N/A

**Release note**:

```release-note
* Update interface implementation of ``GEventMessage``.
* Add threadsafety when adding a message to the success channel from ``GPSEventConsumer``.
* Add flow control when consuming from Pub/Sub.
* Remove date validation in schemas.
```


docs/config-service.rst
```diff
@@ -62,6 +62,11 @@ All configuration options above in the general ``[gcp]`` may be used here. Addit
 
     For more information on Google Pub/Sub subscriptions, please see `Google's docs on managing subscriptions <subscriptions>`_.
 
+.. option:: max_messages=INT
+
+    `Optional`: Number of Pub/Sub messages to process at a time. Defaults to 25.
```
econchick: The default I feel is pretty conservative; but we can always update it if we want.

tests/unit/plugins/service/test_event_consumer.py
```diff
@@ -104,6 +104,22 @@ def test_event_msg_append_to_history(mocker, pubsub_msg, raw_msg_data):
...
 
 
+def test_event_msg_update_phase(mocker, pubsub_msg, raw_msg_data):
+    """Update phase of the message and write to history log."""
+    mocker.patch(DATETIME_PATCH, conftest.MockDatetime)
+    msg = event_consumer.GEventMessage(pubsub_msg, raw_msg_data)
+
+    msg.update_phase('emo')
+
+    assert 'emo' == msg.phase
```
matt0bi: What do you think of using a legit looking phase here? I'm thinking of the person who isn't familiar with Gordon and is trying to grok how phases work and this could serve as a clue.

src/gordon_gcp/plugins/service/event_consumer.py
```diff
@@ -372,7 +372,9 @@ def _get_and_validate_pubsub_msg_schema(self, pubsub_msg_data):
 
         msg_logger.debug(f'Adding message to the success channel.')
         event_msg.update_phase(self.start_phase)
-        await self.success_channel.put(event_msg)
+
+        coro = self.success_channel.put(event_msg)
+        asyncio.run_coroutine_threadsafe(coro, self._loop)
```
matt0bi: :tada:

docs/config-service.rst
matt0bi: Do you know what's the max number of messages Gordon can handle? Is it related to the thread limit?

tests/unit/plugins/service/test_event_consumer.py
econchick: ha, I used this for other phase testing. I'd hope the prose docs and docstrings would help?

docs/config-service.rst
econchick: I'm not sure; it can handle 25, but I haven't load tested. It's certainly related to thread limit, but also related to async tasks, too. As I've heard, performance starts to degrade around 50 threaded requests, which I'm not sure translates to async http calls. I feel like 25 is safe to start off at and can turn up from there. What do you think?

tests/unit/plugins/service/test_event_consumer.py
econchick: (sorry this resp ^ was unclear) I didn't originally think it would be misleading but I had forgotten that some people read tests as docs. If changes are wanted, then I think a separate PR should be made to clean up the sprinkled bits of humor like this.

matt0bi: Ah, you're right about this being the case elsewhere. I agree that this should be a separate PR, but I'd be curious to hear @descent3me and @smstone weight if this is worth pursuing.

docs/config-service.rst
matt0bi: Yeah, 25 sounds like a good place to start and I think we'll smoke out a better number once we have Gordon running. Thanks for shedding some light on this!

src/gordon_gcp/plugins/service/event_consumer.py
```diff
@@ -365,7 +376,10 @@ def _get_and_validate_pubsub_msg_schema(self, pubsub_msg_data):
             pubsub_msg, event_msg_data, schema)
 
         msg_logger.debug(f'Adding message to the success channel.')
-        await self.success_channel.put(event_msg)
+        event_msg.update_phase(self.start_phase)
```
descent3me: Double-pasting this comment since it got hidden when I commented commit by commit.

This is my bad. I actually meant to pass start_phase on line 327/338 [(if the link works](https://github.com/spotify/gordon-gcp/pull/37/files#diff-ba365bd6efd92a959016bfbec05ccc0fR338)) so that it's initialized with the phase, but it looks like I only added it to the log message instead. I think it makes more sense to initialize it with a start phase rather than update it with one.

matt0bi:
:+1:

src/gordon_gcp/plugins/service/event_consumer.py
econchick: Ah I like that better - I'll change it

econchick: @descent3me is this what you had in mind? (updated the commit message too)

src/gordon_gcp/plugins/service/event_consumer.py
```diff
@@ -324,7 +335,7 @@ def __init__(self, config, success_channel, error_channel, metrics,
 
     def _create_gevent_msg(self, pubsub_msg, data, schema):
         log_entry = f'Created a "{schema}" message.'
-        msg = GEventMessage(pubsub_msg, data)
+        msg = GEventMessage(pubsub_msg, data, phase=self.start_phase)
```
descent3me: yup 👍

descent3me: But actually now that I think about maybe this message handling should be in the GEventMessage emit too?

requirements.txt
```diff
@@ -4,7 +4,7 @@ attrs==17.4.0
 google-api-core==0.1.4
 google-auth==1.3.0
 google-cloud-pubsub==0.30.1
-gordon-dns==0.0.1.dev3
+gordon-dns==0.0.1.dev5
```
smstone: should be `0.0.1.dev6` now

econchick: Good catch - I'll fix it for the repo/dev environment, but it won't effect the packaging + our internal deployment.

PR #42, opened 2018-06-22
Extract rrsets properly and set params as optional parameter when calling http.get_all

smstone:
This PR addresses to runtime issues, where we were not properly extracting `rrsets` from the Google Cloud DNS rrset endpoint. The other addresses an issue where `http.get_all` expects there to always be `params` dict passed in.

Fixes #

**Release note**:

```release-note
Extract response rrsets properly
Make params optional when calling http.get_all
```

@spotify/alf

### DIFF
--- a/src/gordon_gcp/plugins/service/gdns_publisher.py
+++ b/src/gordon_gcp/plugins/service/gdns_publisher.py
@@ -245,16 +245,15 @@ class GDNSPublisher:
         Returns:
             dict containing original additions plus deletions.
         """
-        resp = await self.http_client.get_all(resource_records_url)
-        rrsets = resp['rrsets']
-
         record_name = changes['additions'][0]['name']
         record_type = changes['additions'][0]['type']
 
-        for rec in rrsets:
-            if rec['name'] == record_name and rec['type'] == record_type:
-                changes['deletions'] = [rec]
-                break
+        response = await self.http_client.get_all(resource_records_url)
+        for resp in response:
+            for rec in resp['rrsets']:
+                if rec['name'] == record_name and rec['type'] == record_type:
+                    changes['deletions'] = [rec]
+                    break
 
         return changes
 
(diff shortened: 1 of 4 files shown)


src/gordon_gcp/plugins/service/gdns_publisher.py
```diff
@@ -245,16 +245,15 @@ def _format_resource_record_changes(self, action, resource_record):
...
         record_name = changes['additions'][0]['name']
         record_type = changes['additions'][0]['type']
 
-        for rec in rrsets:
-            if rec['name'] == record_name and rec['type'] == record_type:
-                changes['deletions'] = [rec]
-                break
+        response = await self.http_client.get_all(resource_records_url)
+        for resp in response:
+            for rec in resp['rrsets']:
```
econchick: Hmm I wonder if it'd be better to change the `http.get_all` api instead? It seems like something reasonable to expect - that the response isn't paginated. Thoughts from @matt0bi @descent3me ?

matt0bi: I like the idea of flattening the result of `http.get_all` :+1:. Would it mean a lot of changes for in other parts of the code?

descent3me: I agree that it wouldn't be unreasonble to get something like that back from something called `get_all`, but I'm not sure how consistent different google endpoints are in that aspect, i.e. will there always be some named list (in this case 'rrsets') that can be combined? How do you identify it if you don't know what the key will be? Are there totals that need to be updated? It might be simple, but it'd require some investigation, so I'm fine not making the change, since I think getting multiple full response objects from something called `get_all` is also reasonable.

smstone:
@spotify/alf any more feedback?

smstone:
hey @spotify/alf any other comments/feedback?

descent3me:
I got nothing else 👍

PR #43, opened 2018-06-22
Janitor: Skip project if listing instances fails.

smstone:
This PR skips projects where a HTTP error is raised when attempting to list the project's instances. 

This is needed due because when a GCP project is deleted, the instances endpoint will return a 404. 

Fixes #

**Release note**:

```release-note
Janitor: Skip project if listing instances fails.
```
@spotify/alf

### DIFF
--- a/src/gordon_gcp/plugins/janitor/authority.py
+++ b/src/gordon_gcp/plugins/janitor/authority.py
@@ -143,8 +143,14 @@ class GCEAuthority:
     async def _get_instances(self, projects):
         instance_filter = self.config.get('instance_filter')
         for project in projects:
-            yield await self.gce_client.list_instances(
-                project, instance_filter=instance_filter)
+            try:
+                yield await self.gce_client.list_instances(
+                    project, instance_filter=instance_filter)
+            except exceptions.GCPHTTPError as e:
+                logging.warn(
+                    f'Could not fetch instance list for project {project}:'
+                    f'{e}. Skipping this project.')
+                continue
 
     def _create_instance_rrset(self, instance):
         ip = instance['networkInterfaces'][0]['accessConfigs'][0]['natIP']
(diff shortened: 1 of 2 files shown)


src/gordon_gcp/plugins/janitor/authority.py
```diff
@@ -143,8 +143,14 @@ def __init__(self, config, crm_client, gce_client, rrset_channel=None,
     async def _get_instances(self, projects):
         instance_filter = self.config.get('instance_filter')
         for project in projects:
-            yield await self.gce_client.list_instances(
-                project, instance_filter=instance_filter)
+            try:
+                yield await self.gce_client.list_instances(
+                    project, instance_filter=instance_filter)
+            except exceptions.GCPHTTPError as e:
```
matt0bi: Should we differentiate between 4xx and 5xx responses? If we get the occasional 404, there's probably nothing to worry about, but if every response is a 502 "timeout", I'm not sure how we'd know that something's wrong? Unless I'm overthinking this.

smstone: hm, good point - the current setup doesn't make it too easy to extract which status code is returned.

descent3me: I can't really say what we'd do different. We'd still skip, right?

smstone: yeah, i believe so. we wouldn't want the janitor to bail out (much like zonextgen) on an failed instance list on a project. i think one thing we could do is add some metrics around this and alert accordingly if there are many failures when calling out to 3rd party APIs? we have a ticket to track this already.

matt0bi: > i think one thing we could do is add some metrics around this and alert accordingly if there are many failures when calling out to 3rd party APIs? we have a ticket to track this already.

I was going to suggest this, but since we already have a ticket :+1:

PR #45, opened 2018-06-28
Remove the GPThread instance once done

econchick:

Since we didn't ever remove `GPThread` instances from the `GEventConsumer` dict, threads were never released/gc'ed (we think), causing an every-growing amount of open file descriptors. 

Fixes #

**Release note**:

```release-note
Clean up GPThread instances once done.
```

@spotify/alf


ocsw:
+1
Are we adding thread.join() in another PR?

matt0bi:
:+1:, let's try this.

I'm not 100% confident this will solve the problem, but it should at least allow Gordon to run for longer than a few hours/days :crossed_fingers:.

PR #49, opened 2018-07-06
Filter out root SOA/NS records in reconciler

ocsw:
I realized that if we pull every record from GDNS in the authority, it will include SOA and NS. If those are present in the GHE input, they will be ignored, and it'll try to add them every time. If not, and we have deletion, we may try to delete them (which will either fail or blow up). So we need to remove SOA and NS records from the comparison in the authority. Note that this only applies to NS records for the root of the zone; Google seems to accept others in my testing.

Fixes #

**Release note**:

```release-note
Filtered out SOA/NS from janitor reconciliation.
```

### DIFF
--- a/src/gordon_gcp/plugins/janitor/reconciler.py
+++ b/src/gordon_gcp/plugins/janitor/reconciler.py
@@ -240,6 +240,32 @@ class GDNSReconciler:
 
         return zone, rrsets
 
+    def _remove_soa_and_root_ns(self, zone, rrsets):
+        """Given a list of GCPResourceRecordSets, remove SOA and root NS.
+
+        This is necessary because Google won't let you override its SOA and root
+        NS records.
+
+        Returns a new list (doesn't modify in place).
+
+        Args:
+            zone (str): The domain name the rrsets are for, including trailing
+                dot.
+            rrsets (list[GCPResourceRecordSet]): The GCPResourceRecordSets to
+                filter.
+
+        Returns:
+            list[GCPResourceRecordSet]: The filtered GCPResourceRecordSets.
+        """
+        clean_rrsets = []
+        for rrset in rrsets:
+            if rrset.type == 'SOA':
+                continue
+            if rrset.type == 'NS' and rrset.name == zone:
+                continue
+            clean_rrsets.append(rrset)
+        return clean_rrsets
+
     async def validate_rrsets_by_zone(self, zone, rrsets):
         """Given a zone, validate current versus desired rrsets.
 
@@ -264,13 +290,13 @@ class GDNSReconciler:
         missing_rrsets = [
             rs for rs in desired_rrsets if rs not in actual_rrsets
         ]
+        # don't try to add root SOA/NS records
+        missing_rrsets = self._remove_soa_and_root_ns(zone, missing_rrsets)
+
         msg = (f'[{zone}] Processed {len(actual_rrsets)} rrset messages '
                f'and found {len(missing_rrsets)} missing rrsets.')
         logging.info(msg)
-        # TODO (lynn): emit or incr metric of missing_rrsets by zone once
-        #              aioshumway is released
-        # TODO: (FEATURE): have separate metrics for additions and deletions
-        await self.publish_change_messages(missing_rrsets, action='additions')
+        return missing_rrsets
 
     async def run(self):
         """Start consuming from :obj:`rrset_channel`.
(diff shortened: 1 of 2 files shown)


ocsw:
@spotify/alf ok?

src/gordon_gcp/plugins/janitor/reconciler.py
```diff
@@ -240,6 +240,33 @@ def _parse_rrset_message(self, message):
 
         return zone, rrsets
 
+    @staticmethod
```
smstone: what's the reasoning for using a static method here?

src/gordon_gcp/plugins/janitor/reconciler.py
```diff
@@ -240,6 +240,33 @@ def _parse_rrset_message(self, message):
...
+        Returns a new list (doesn't modify in place).
+
+        Args:
+            zone (str): The domain name the rrsets are for, including trailing
+                dot.
+            rrsets (list[GCPResourceRecordSet]): The GCPResourceRecordSets to
+                filter.
+
+        Returns:
+            (list[GCPResourceRecordSet]): The filtered GCPResourceRecordSets.
```
smstone: do we need the parenthesis around this?

src/gordon_gcp/plugins/janitor/reconciler.py
```diff
@@ -264,6 +291,10 @@ def _parse_rrset_message(self, message):
         missing_rrsets = [
             rs for rs in desired_rrsets if rs not in actual_rrsets
         ]
+        # don't try to add root SOA/NS records
+        missing_rrsets = GDNSReconciler._remove_soa_and_root_ns(
```
smstone: i'm wondering if these calls should live outside of `validate_rrsets_by_zone`, as they don't really relate to validating the rrsets by zone. perhaps just in the run method before calling `validate_rrsets_by_zone`?

src/gordon_gcp/plugins/janitor/reconciler.py
ocsw: Well, it doesn't need anything from the instance, so...

src/gordon_gcp/plugins/janitor/reconciler.py
ocsw: Hm.  I guess not.  I seem to be remembering the style guide wrong.

src/gordon_gcp/plugins/janitor/reconciler.py
ocsw: I mean, that's kind of a problem with the function to begin with, since it publishes the changes, not just validates them.  I can split it up, but I think in that case the filtering should be in the validator and the publish_change_messages calls should be in run().

src/gordon_gcp/plugins/janitor/reconciler.py
smstone: hm, maybe i am missing something - but since it's usage from `validate_rrsets_by_zone` is in the class anyway, why not just call it by doing `_remove_soa_and_root_ns`?

ocsw: The call would be either GDNSReconciler._remove_soa_and_root_ns (static method) or self._remove_soa_and_root_ns (instance method).  It's not strictly required to make this a static method, but since it doesn't need anything from the instance, that's what I generally do.  That way it can, in theory, be used by something else without instantiating the class.  If you think it's a bad idea I can change it.

ocsw: Oh this is interesting.  Turns out the decorator makes it possible to call the method from the instance, and isn't necessary otherwise.  So I could change the calls without changing the method.  I do like to make it clear that it's a static method in how I call it though.

ocsw: @smstone pointed out that it's also private.  Meh, made it instance.

ocsw:
Ok, I split the function up a bit, which involved changing the tests...

@spotify/alf can I get another amen?

ocsw:
(tests are failing because of the addition of 'kind' to the rrset class - I think the proper way to fix it is to add a default there)

ocsw:
(I fixed the test problem by adding kind fields.  I think all our uses should actually be covered, but I also still think a default is a good idea.)

src/gordon_gcp/plugins/janitor/reconciler.py
```diff
@@ -264,13 +290,13 @@ def _parse_rrset_message(self, message):
         missing_rrsets = [
             rs for rs in desired_rrsets if rs not in actual_rrsets
         ]
+        # don't try to add root SOA/NS records
+        missing_rrsets = self._remove_soa_and_root_ns(zone, missing_rrsets)
```
descent3me: I missed this PR, but maybe this should happen against `actual_rrsets` instead, since these are actual rrsets that we don't care about it. Obviously if someone wants to create NS or SOA (i.e. in `desired_rrsets`) that won't work, but maybe we should be hard failing in that case, instead of silently accepting but ignoring them.

PR #52, opened 2018-07-09
Add kind attribute to GCPResourceRecordSet

smstone:
This adds a `kind` attribute to `gcp.GCPResourceRecordSet`. This is part of the RRset that returned by the Google Cloud DNS resourceRecordSets API https://cloud.google.com/dns/api/v1/resourceRecordSets#resource

A type error is currently raised without this attribute.

Fixes #

**Release note**:

```release-note

```
@spotify/alf

### DIFF
--- a/src/gordon_gcp/clients/gdns.py
+++ b/src/gordon_gcp/clients/gdns.py
@@ -61,6 +61,8 @@ class GCPResourceRecordSet:
 
     Args:
         name (str): Name/label.
+        kind (str): ID for what kind of GCP resource this is. For example
+            'dns#resourceRecordSet'
         type (str): Record type (see `Google's supported records
             <https://cloud.google.com/dns/overview#supported_dns_record_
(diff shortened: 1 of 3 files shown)


src/gordon_gcp/clients/gdns.py
```diff
@@ -61,6 +61,8 @@ class GCPResourceRecordSet:
 
     Args:
         name (str): Name/label.
+        kind (str): ID for what kind of DNS resource this is. For example
```
RochesterinNYC: Isn't the `kind` more of a "what kind of GCP resource" the resource is ? (As in, it's a GCP thing and not a Google DNS thing).

smstone: ah yeah - i'll switch it to GCP. I put 'DNS' since it's only dealing with the Google DNS Resource record set API, but having GCP resource here makes sense too.

PR #56, opened 2018-07-10
Add populating of 'resourceRecords' for GEventMessages that have the 'deletions' action

RochesterinNYC:

Right now, when parsing audit log pub/sub messages, the consumer sets the `resourceRecords` of the GEventMessage to an empty list, and the enricher assumes any that this empty`resourceRecords` needs to be enriched. However, that doesn't necessarily work for `deletions` actions, as when the enricher tries to find the related instance information from the GCP Instances API, the instance is already gone/deleted and will hence 404. 

This PR adds the ability to populate the `resourceRecords` for GEventMessages that have the 'deletions' action via querying the Google Cloud DNS API for the records that Gordon initially created for the GCE host. Currently, this would only be the A record.

The `name` and `type` parameters for the Google DNS API List endpoint are used to and only get the results/records that Gordon initially created for the GCE host (currently only the A record).

**Release note**:

```release-note
Add populating of 'resourceRecords' for GEventMessages that have the 'deletions' action
```

### DIFF
--- a/gordon.toml.example
+++ b/gordon.toml.example
@@ -7,10 +7,11 @@ handlers = ["syslog"]
 # Plugin Config
 [gcp]
 project = "my-gcp-project"
+dns_zone = "example.com."
+managed_zone = "example-com"
 
 [gcp.enricher]
 keyfile = "/path/to/keyfiles/compute.json"
-dns_zone = "example.com."
 
 [gcp.event_consumer]
 keyfile = "/path/to/keyfiles/pubsub.json"
@@ -23,5 +24,3 @@ project = "my-dns-project"
 keyfile = "/path/to/keyfiles/dns.json"
 publish_wait_timeout = 10
 default_ttl = 300
-dns_zone = "example.com."
-managed_zone = "example-zone"
--- a/src/gordon_gcp/plugins/service/enricher.py
+++ b/src/gordon_gcp/plugins/service/enricher.py
@@ -266,6 +266,7 @@ class GCEEnricher:
             instance_resource_url = event_message.data['resourceName']
             records = await self._get_matching_records_for_deletion(
                 instance_resource_url)
+
         msg_logger.debug(f'Enriched with resource record(s): {records}')
         event_message.data['resourceRecords'].extend(records)
         msg = (f"Enriched msg with {len(event_message.data['resourceRecords'])}"
--- a/src/gordon_gcp/plugins/service/enricher.py
+++ b/src/gordon_gcp/plugins/service/enricher.py
@@ -215,25 +215,24 @@ class GCEEnricher:
                 fqdn, external_ip, default_ttl, msg_logger)
         ]
 
-    async def _get_matching_records(self, instance_resource_url):
+    # The types of records that this returns for deletion should match the
+    # types of records that are created above in the _create_rrecords method
+    async def _get_matching_records_for_deletion(self, instance_resource_url):
         # Get the fqdn of the instance
         instance_name = instance_resource_url.split('/')[-1]
         fqdn = self._get_fqdn(instance_name)
 
-        # Get all Google DNS records for the managed zone
+        # Get the Google DNS A records for managed zone with name matching the
+        # instance fqdn
         project = self.config['project']
         managed_zone = self.config['managed_zone']
         resource_records_url = GCEEnricher.RESOURCE_RECORDS_ENDPOINT.format(
             project=project, managedZone=managed_zone)
-        response = await self._http_client.get_all(resource_records_url)
 
-        # Get the Google DNS A records with name matching the instance fqdn
-        matching_records = []
-        for resp in response:
-            for rec in resp['rrsets']:
-                if rec['name'] == fqdn:
-                    matching_records.append(rec)
-        return matching_records
+        search_params = {'name': fqdn, 'type': 'A'}
+        response = await self._http_client.get_json(
+            resource_records_url, params=search_params)
+        return response['rrsets']
 
     async def handle_message(self, event_message):
         """
@@ -265,7 +264,8 @@ class GCEEnricher:
                 event_message.data, instance_data, msg_logger)
         elif event_message.data['action'] == 'deletions':
             instance_resource_url = event_message.data['resourceName']
-            records = await self._get_matching_records(instance_resource_url)
+            records = await self._get_matching_records_for_deletion(
+                instance_resource_url)
         msg_logger.debug(f'Enriched with resource record(s): {records}')
         event_message.data['resourceRecords'].extend(records)
         msg = (f"Enriched msg with {len(event_message.data['resourceRecords'])}"
--- a/src/gordon_gcp/plugins/service/enricher.py
+++ b/src/gordon_gcp/plugins/service/enricher.py
@@ -80,6 +80,22 @@ class GCEEnricherBuilder:
             self.config['retries'] = 5
         return []
 
+    def _validate_managed_zone(self):
+        msg = []
+        if not self.config.get('managed_zone'):
+            msg.append('The name of the Google Cloud DNS managed zone is '
+                       'required to correctly delete A records for deleted '
+                       'instances.')
+        return msg
+
+    def _validate_project(self):
+        msg = []
+        if not self.config.get('project'):
+            msg.append('The GCP project that contains the Google Cloud DNS '
+                       'managed zone is required to correctly delete A records '
+                       'for deleted instances.')
+        return msg
+
     def _call_validators(self):
         """Actually run all the validations.
 
@@ -90,6 +106,8 @@ class GCEEnricherBuilder:
         msg.extend(self._validate_keyfile())
         msg.extend(self._validate_dns_zone())
         msg.extend(self._validate_retries())
+        msg.extend(self._validate_project())
+        msg.extend(self._validate_managed_zone())
         return msg
 
     def _validate_config(self):
@@ -126,6 +144,8 @@ class GCEEnricher:
             the GCE API.
     """
     phase = 'enrich'
+    RESOURCE_RECORDS_ENDPOINT = ('https://www.googleapis.com/dns/v1/projects/'
+                                 '{project}/managedZones/{managedZone}/rrsets')
 
     def __init__(self, config, metrics, http_client, **kwargs):
         self.config = config
@@ -195,6 +215,26 @@ class GCEEnricher:
                 fqdn, external_ip, default_ttl, msg_logger)
         ]
 
+    async def _get_matching_records(self, instance_resource_url):
+        # Get the fqdn of the instance
+        instance_name = instance_resource_url.split('/')[-1]
+        fqdn = self._get_fqdn(instance_name)
+
+        # Get all Google DNS records for the managed zone
+        project = self.config['project']
+        managed_zone = self.config['managed_zone']
+        resource_records_url = GCEEnricher.RESOURCE_RECORDS_ENDPOINT.format(
+            project=project, managedZone=managed_zone)
+        response = await self._http_client.get_all(resource_records_url)
+
+        # Get the Google DNS A records with name matching the instance fqdn
+        matching_records = []
+        for resp in response:
+            for rec in resp['rrsets']:
+                if rec['name'] == fqdn:
+                    matching_records.append(rec)
+        return matching_records
+
     async def handle_message(self, event_message):
         """
         Enrich message with extra context and send it to the publisher.
@@ -208,7 +248,6 @@ class GCEEnricher:
             event_message (.GEventMessage): message requiring
                 additional information.
         """
-
         # if the message has resource records, assume it has all info
         # it needs to be published, and therefore is already enriched
         if event_message.data['resourceRecords']:
@@ -219,14 +258,16 @@ class GCEEnricher:
         msg_logger = _utils.GEventMessageLogger(
             self._logger, {'msg_id': event_message.msg_id})
 
-        instance_data = await self._poll_for_instance_data(
-            event_message.data['resourceName'], msg_logger)
-        records = await self._create_rrecords(
-            event_message.data, instance_data, msg_logger)
-
+        if event_message.data['action'] == 'additions':
+            instance_data = await self._poll_for_instance_data(
+                event_message.data['resourceName'], msg_logger)
+            records = await self._create_rrecords(
+                event_message.data, instance_data, msg_logger)
+        elif event_message.data['action'] == 'deletions':
+            instance_resource_url = event_message.data['resourceName']
+            records = await self._get_matching_records(instance_resource_url)
         msg_logger.debug(f'Enriched with resource record(s): {records}')
         event_message.data['resourceRecords'].extend(records)
-
-        added_records = event_message.data['resourceRecords']
-        msg = f'Enriched msg with {len(added_records)} resource record(s).'
+        msg = (f"Enriched msg with {len(event_message.data['resourceRecords'])}"
+               " resource record(s).")
         event_message.append_to_history(msg, self.phase)
--- a/tests/unit/plugins/service/test_enricher.py
+++ b/tests/unit/plugins/service/test_enricher.py
@@ -22,9 +22,9 @@ from gordon_gcp.plugins.service import enricher
 
 
 @pytest.fixture
-def instance_data(audit_log_data):
+def instance_data(creation_audit_log_data):
     return {
-        'name': audit_log_data['resourceName'].split('/')[-1],
+        'name': creation_audit_log_data['resourceName'].split('/')[-1],
         'networkInterfaces': [{
             'accessConfigs': [{
                 'natIP': '127.99.199.27'
@@ -77,12 +77,12 @@ def gce_enricher(config, metrics, mock_http_client):
 
 
 @pytest.mark.asyncio
-async def test_handle_message_doesnt_need_enriching(mocker, caplog, gevent_msg,
-                                                    gce_enricher):
-    gevent_msg.data['resourceRecords'] = [mocker.Mock()]
+async def test_handle_message_doesnt_need_enriching(
+        mocker, caplog, additions_gevent_msg, gce_enricher):
+    additions_gevent_msg.data['resourceRecords'] = [mocker.Mock()]
     expected_msg = 'Message already enriched, skipping phase.'
-    await gce_enricher.handle_message(gevent_msg)
-    assert expected_msg == gevent_msg.history_log[0]['message']
+    await gce_enricher.handle_message(additions_gevent_msg)
+    assert expected_msg == additions_gevent_msg.history_log[0]['message']
     assert 0 == len(caplog.records)
 
 
@@ -90,9 +90,9 @@ async def test_handle_message_doesnt_need_enriching(mocker, caplog, gevent_msg,
     (0, 3),
     (2, 5)])
 @pytest.mark.asyncio
-async def test_handle_message_event_msg(
-        mocker, config, gevent_msg, mock_async_sleep, gce_enricher, caplog,
-        instance_data, sleep_calls, logs_logged):
+async def test_handle_message_event_msg_additions(
+        mocker, config, additions_gevent_msg, mock_async_sleep, gce_enricher,
+        caplog, instance_data, sleep_calls, logs_logged):
     """Successfully enrich event message."""
     instance_mocked_data = []
     for i in range(sleep_calls):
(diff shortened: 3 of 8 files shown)


tests/unit/plugins/service/test_enricher.py
```diff
@@ -127,20 +162,52 @@ def gce_enricher(config, metrics, mock_http_client):
...
+        response, sleep_calls, logs_logged, err_msg):
+    """Raise error while enriching event message with additions."""
     gce_enricher._http_client._get_json_mock.side_effect = response
-    gevent_msg.phase = 'enrich'
 
     with pytest.raises(exceptions.GCPGordonError) as e:
-        await gce_enricher.handle_message(gevent_msg)
+        await gce_enricher.handle_message(additions_gevent_msg)
 
-    assert 'enrich' == gevent_msg.phase
```
RochesterinNYC: I removed this assertion as it didn't seem to really test anything? Since the test manually set the `gevent_msg.phase` value via `gevent_msg.phase = 'enrich'` on line 135.

src/gordon_gcp/plugins/service/enricher.py
```diff
@@ -195,6 +197,24 @@ def _get_internal_ip(self, instance_data):
                 fqdn, external_ip, default_ttl, msg_logger)
         ]
 
+    async def _get_matching_records(self, instance_resource_url):
+        # Get all Google DNS records for the zone
+        project = self.config.get('project')
+        managed_zone = self.config['managed_zone']
```
RochesterinNYC: I might be misunderstanding the purpose of the managed_zone config and how it relates separately to the publisher, enricher, janitor, etc. How many publishers, enrichers, etc. are there configured for each managed zone? Can the enricher have `managed_zone` as part of its config?

RochesterinNYC: Moved the `managed_zone` attribute up to the `gcp` level in the config to illustrate where the managed_zone configuration might be.

src/gordon_gcp/plugins/service/enricher.py
```diff
@@ -195,6 +197,27 @@ def _get_internal_ip(self, instance_data):
                 fqdn, external_ip, default_ttl, msg_logger)
         ]
 
+    async def _get_matching_records(self, instance_resource_url):
+        # Get the fqdn of the instance
+        instance_name = instance_resource_url.split('/')[-1]
+        fqdn = self._get_fqdn(instance_name)
+
+        # Get all Google DNS records for the managed zone
```
descent3me: You might need to update the comments after the latest changeset

src/gordon_gcp/plugins/service/enricher.py
```diff
@@ -195,6 +197,27 @@ def _get_internal_ip(self, instance_data):
...
+        project = self.config.get('project')
```
ocsw: Are we sure this isn't required?  If it is, it should be validated in the builder, and accessed directly here.

src/gordon_gcp/plugins/service/enricher.py
```diff
@@ -195,6 +197,27 @@ def _get_internal_ip(self, instance_data):
...
+        managed_zone = self.config['managed_zone']
+        resource_records_url = GCEEnricher.RESOURCE_RECORDS_ENDPOINT.format(
+            project=project, managedZone=managed_zone)
+        search_params = {'name': fqdn}
+        response = await self._http_client.get_all(
+            resource_records_url, params=search_params)
+
+        # Get the Google DNS records with name matching the instance fqdn
+        matching_records = []
+        for resp in response:
```
ocsw: This may not need to be a loop - there should only be at most one response (I think?)

tests/unit/plugins/service/test_enricher.py
```diff
@@ -77,39 +110,41 @@ def gce_enricher(config, metrics, mock_http_client):
...
     gce_enricher._http_client._get_json_mock.side_effect = instance_mocked_data
 
-    await gce_enricher.handle_message(gevent_msg)
+    await gce_enricher.handle_message(additions_gevent_msg)
 
     expected_history_msg = 'Enriched msg with 1 resource record(s).'
-    assert expected_history_msg == gevent_msg.history_log[0]['message']
-    assert 'enrich' == gevent_msg.history_log[0]['plugin']
-    assert 1 == len(gevent_msg.data['resourceRecords'])
+    assert expected_history_msg == \
```
ocsw: use () here for line continuation

tests/unit/plugins/service/test_enricher.py
```diff
@@ -127,20 +162,52 @@ def gce_enricher(config, metrics, mock_http_client):
...
-async def test_handle_message_event_msg_failures(
-        mocker, gevent_msg, mock_async_sleep, gce_enricher, caplog, response,
-        sleep_calls, logs_logged, err_msg):
-    """Raise error while enriching event message."""
+async def test_handle_message_event_msg_additions_failures(
+        mocker, additions_gevent_msg, mock_async_sleep, gce_enricher, caplog,
+        response, sleep_calls, logs_logged, err_msg):
+    """Raise error while enriching event message with additions."""
     gce_enricher._http_client._get_json_mock.side_effect = response
-    gevent_msg.phase = 'enrich'
```
ocsw: Are you sure this isn't required?

gordon.toml.example
```diff
@@ -7,6 +7,7 @@ handlers = ["syslog"]
 # Plugin Config
 [gcp]
 project = "my-gcp-project"
+managed_zone = "example-zone"
```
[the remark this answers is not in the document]
ocsw: Also, I would make this example-com to match our internal practice.

tests/unit/plugins/service/test_enricher.py
```diff
@@ -33,12 +33,42 @@ def instance_data(audit_log_data):
...
+    ]
+
+
 @pytest.fixture
 def config(fake_keyfile):
     return {
         'keyfile': fake_keyfile,
         'scopes': [],
         'dns_zone': 'example.com.',
+        'managed_zone': 'example-zone',
```
ocsw: I would make this example-com to match our internal practice.

tests/unit/plugins/service/test_enricher.py
```diff
@@ -127,20 +162,52 @@ def gce_enricher(config, metrics, mock_http_client):
...
+async def test_handle_message_event_msg_deletions(
+        mocker, config, deletions_gevent_msg, gce_enricher, caplog,
+        records_data):
+    """Successfully enrich event message with records to delete."""
+    gce_enricher._http_client._get_all_mock.return_value = records_data
+
+    await gce_enricher.handle_message(deletions_gevent_msg)
+
+    expected_history_msg = 'Enriched msg with 2 resource record(s).'
+    assert expected_history_msg == \
```
ocsw: use () for continuation

tests/unit/plugins/service/test_enricher.py
```diff
@@ -127,20 +162,52 @@ def gce_enricher(config, metrics, mock_http_client):
...
+        deletions_gevent_msg.history_log[0]['message']
+    assert 'enrich' == deletions_gevent_msg.history_log[0]['plugin']
```
ocsw: not enricher?

tests/unit/plugins/service/test_enricher.py
```diff
@@ -127,20 +162,52 @@ def gce_enricher(config, metrics, mock_http_client):
...
+    assert 2 == len(deletions_gevent_msg.data['resourceRecords'])
```
ocsw: not necessary if you test for equality (below)

tests/unit/plugins/service/test_enricher.py
```diff
@@ -127,20 +162,52 @@ def gce_enricher(config, metrics, mock_http_client):
...
+    fqdn = ('.'.join([deletions_gevent_msg.data['resourceName'].split('/')[-1],
+            config['dns_zone']]))
+    expected_rrecords = [{
```
ocsw: Can this be derived from records_data without duplication?

tests/unit/plugins/service/test_enricher.py
```diff
@@ -127,20 +162,52 @@ def gce_enricher(config, metrics, mock_http_client):
...
     assert sleep_calls == mock_async_sleep.call_count
     assert logs_logged == len(caplog.records)
     expected_msg = ('Could not get necessary information for '
                     'projects/123456789101/zones/us-central1-c/'
                     f'instances/an-instance-name-b34c: {err_msg}')
     assert e.match(expected_msg)
+
+
+@pytest.mark.asyncio
+async def test_handle_message_event_msg_deletions(
```
ocsw: How about a test for if there are no records to enrich with?

src/gordon_gcp/plugins/service/enricher.py
```diff
@@ -219,14 +242,16 @@ def _get_internal_ip(self, instance_data):
...
+            instance_resource_url = event_message.data['resourceName']
+            records = await self._get_matching_records(instance_resource_url)
         msg_logger.debug(f'Enriched with resource record(s): {records}')
         event_message.data['resourceRecords'].extend(records)
-
-        added_records = event_message.data['resourceRecords']
-        msg = f'Enriched msg with {len(added_records)} resource record(s).'
+        msg = (f"Enriched msg with {len(event_message.data['resourceRecords'])}"
+               " resource record(s).")
         event_message.append_to_history(msg, self.phase)
```
ocsw: Should we be doing something here if there are zero records?

gordon.toml.example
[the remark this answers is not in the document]
RochesterinNYC: I'm not sure what your first comment means ?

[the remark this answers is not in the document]
ocsw: So there was a PR that made some changes to these sample configs.  I don't remember what it was, exactly, but you should be able to track it down with git blame.  IIRC, I suggested that 'managed_zone' should be next to 'dns_domain' to make it harder to forget to change one of them when changing the other.

src/gordon_gcp/plugins/service/enricher.py
RochesterinNYC: I don't think there's any harm in keeping this as a loop, since there's no guarantee that the user has not added a bunch of records for that fqdn and that the response from `get_all` could include multiple pages.

Though on a similar note, do we want to ensure that the behavior for Gordon on deletion of a GCE instance **only** deletes records that Gordon itself created (in this case it would be just the A record), or will both delete the A record that Gordon itself created but also any records in Google Cloud DNS with the instance's fqdn as the record name?

example:
- `test-host-fdsa2.example.com` GCE instance is created
-  Gordon creates the A record for `test-host-fdsa2.example.com`
- someone puts into Cloud DNS a bunch of `TXT` records for `test-host-fdsa2.example.com`
- `test-host-fdsa2.example.com` GCE instance is deleted
- Which of these should the behavior for Gordon be:
   - Delete just the A record for `test-host-fdsa2.example.com`
   - Delete all the records for `test.example.com` (A record + all TXT records)

descent3me: I would expect Gordon to only delete records that it creates.

tests/unit/plugins/service/test_enricher.py
RochesterinNYC: I think it should be `enricher`. We set it to `enrich` here: https://github.com/spotify/gordon-gcp/blob/07a39c411ba0fc79f7407a6372ed7f901f479fca/src/gordon_gcp/plugins/service/enricher.py#L128, but since it's `plugin`, `enricher` probably makes more sense. I think `enricher` as a phase is also acceptable nomenclature.

Side note that we're not appending this history bit to the gevent_msg at all in the `gdns_publisher`. Is that on purpose ?

src/gordon_gcp/plugins/service/enricher.py
econchick: @RochesterinNYC if "someone puts into Cloud DNS a bunch of `TXT` records" is done by manual record addition (i.e. via GHE), then I wouldn't expect gordon to delete them without first the user also making a pull request to GHE to remove them.

src/gordon_gcp/plugins/service/enricher.py
```diff
@@ -195,6 +215,26 @@ def _get_internal_ip(self, instance_data):
...
+    async def _get_matching_records(self, instance_resource_url):
+        # Get the fqdn of the instance
+        instance_name = instance_resource_url.split('/')[-1]
+        fqdn = self._get_fqdn(instance_name)
+
+        # Get all Google DNS records for the managed zone
+        project = self.config['project']
+        managed_zone = self.config['managed_zone']
+        resource_records_url = GCEEnricher.RESOURCE_RECORDS_ENDPOINT.format(
+            project=project, managedZone=managed_zone)
```
econchick: Curious - do we plan to use #57 here too?

econchick: d'oh. sorry just got to the next commit 🤦‍♀️

tests/unit/plugins/service/test_enricher.py
RochesterinNYC: The test seems to simply set that value and then check that the value is set. The assertion does not pass without the `gevent_msg.phase = 'enrich'` assignment.

tests/unit/plugins/service/test_enricher.py
ocsw: I don't know, but I think we should.

ocsw: Looks like the phases are verbs, not plugin names, after all.  So enrich is correct unless we want to change them everywhere.

RochesterinNYC: I'll create an issue for it for later. I don't think we should try to stuff it into this PR.

tests/unit/plugins/service/test_enricher.py
ocsw: I did more digging.  The consumer code is supposed to update the gevent_msg with the next phase.  I think the idea was to make sure the input to the test is as real as possible.  I'm not sure if we actually need the assert in that case, though.

However:  It looks like that update is now handled by the router, and should be removed from the consumer?  @econchick ?

descent3me: yeah, that should only happen in the router now

RochesterinNYC: so this line isn't needed anymore if it happens in the router? https://github.com/spotify/gordon-gcp/blob/d3bef0bfbc1943d21fb8896fcb99d0df7cb6e708/src/gordon_gcp/plugins/service/enricher.py#L232

ocsw: No, that line is about history, not phase.  What needs to be removed is update_phase in the consumer.

For the enricher, I'd say leave the assignment line and remove the assert.

descent3me: Well, that just adds a log message... and it looks like that has not been fully integrated into the router yet, which actually should be fixed. Router just calls update phase and it does add a log message but it's just adding a bunch of messages with [plugin=None](https://github.com/spotify/gordon-gcp/blob/develop/src/gordon_gcp/plugins/service/event_consumer.py#L95).  In summary, I think you could remove it from enricher, but you do, you should also be making the logging fix to event_msg (in event_consumer), so that it logs something meaningful.

RochesterinNYC: :+1:

Can one of y'all make the story/issue about the `update_phase` + router stuff? I don't have enough context here.

ocsw: I can.

src/gordon_gcp/plugins/service/enricher.py
RochesterinNYC: Thanks for the input y'all. Switched to using `get_json` instead of `get_all` and now filter for both `name` and `type` of just `A` (since all we're creating are A records at the moment). Added a comment to document the fact that the types of records we return in `_get_matching_records_for_deletion` should match the types created in `_create_rrecords`.

src/gordon_gcp/plugins/service/enricher.py
RochesterinNYC: I've been thinking about this case, and it feels like the best behavior might be to drop/ack the message and appropriately log/record this somewhere?

We might not want to fully treat it as an error as the actual state == the desired state (i.e. we want no A records for the GCE host and the state in Google Cloud DNS is that there are no A records). However, with the logging/recording, we can keep track of how many times this is happening (because this happening too many times indicates something is wrong with the system).

Any examples of this kind of behavior elsewhere in the codebase I can draw inspiration from?

RochesterinNYC:
Updated the PR to address all current comments/feedback.

gordon.toml.example
```diff
@@ -11,6 +11,7 @@ project = "my-gcp-project"
 [gcp.enricher]
 keyfile = "/path/to/keyfiles/compute.json"
 dns_zone = "example.com."
+managed_zone = "example-com"
```
descent3me: I would if instead of adding this here, should both the dns and managed zone be at the gcp level?  I don't know if this was already discussed or if I'm misunderstanding how the config gets parsed, but I can't think of a reason where you'd want the enricher to have a different value than the publisher, and having things in two places is a good way for them to get accidentally out of sync.

src/gordon_gcp/plugins/service/enricher.py
descent3me: I think that sounds ok, but it might add complexity to the system, since right now the assumption is that only failures would cause a message to be dropped from the channel.  We could leave it as a noop for the publisher, and have it just return early if it sees there are no records to publish.

gordon.toml.example
RochesterinNYC: Hm, okay, I had that in this PR originally but had moved it to the current configuration as a result of this comment chain: https://github.com/spotify/gordon-gcp/pull/56#discussion_r201820886

src/gordon_gcp/plugins/service/enricher.py
RochesterinNYC: Updated the PR with the approach of having the Publisher treat GEventMessages without resource records as a no-op and have the enricher not throw an Exception in this case.

gordon.toml.example
RochesterinNYC: Updated it though, as it makes sense to have them together and at a higher level (`gcp`), especially our readthedocs talk more about configuration and overriding values and etc.

descent3me:
+1 from my side

RochesterinNYC:
@ocsw could you check and see if I've addressed all your feedback appropriately ?

src/gordon_gcp/plugins/service/enricher.py
```diff
@@ -195,6 +215,25 @@ def _get_internal_ip(self, instance_data):
...
         ]
 
+    # The types of records that this returns for deletion should match the
+    # types of records that are created above in #_create_rrecords
+    async def _get_matching_records_for_deletion(self, instance_resource_url):
+        # Get the fqdn of the instance
+        instance_name = instance_resource_url.split('/')[-1]
+        fqdn = self._get_fqdn(instance_name)
+
+        # Get the Google DNS records for managed zone with name matching the
```
ocsw: Maybe specify A as well?

src/gordon_gcp/plugins/service/enricher.py
```diff
@@ -195,6 +215,25 @@ def _get_internal_ip(self, instance_data):
                 fqdn, external_ip, default_ttl, msg_logger)
         ]
 
+    # The types of records that this returns for deletion should match the
+    # types of records that are created above in #_create_rrecords
```
ocsw: You've got an extra #.  (and maybe add () for the function name)

tests/unit/plugins/service/test_enricher.py
ocsw: Still needs ()

tests/unit/plugins/service/test_enricher.py
ocsw: I would leave this line in meanwhile.

tests/unit/plugins/service/test_enricher.py
```diff
@@ -127,20 +154,37 @@ def gce_enricher(config, metrics, mock_http_client):
...
+        mocker, config, deletions_gevent_msg, gce_enricher, caplog,
+        records_data, get_json_response_data):
+    """Successfully enrich event message with records to delete."""
+    gce_enricher._http_client._get_json_mock.return_value = \
```
ocsw: use () for continuation

src/gordon_gcp/plugins/service/enricher.py
ocsw: Maybe we should add a test for that in the publisher?

tests/unit/plugins/service/test_enricher.py
RochesterinNYC: It's still in the code, `gevent_msg` has just been renamed to `additions_gevent_msg`.

src/gordon_gcp/plugins/service/enricher.py
RochesterinNYC: Does `test_handle_message_no_resource_records` not cover that ?

PR #57, opened 2018-07-11
Use 'name' and 'type' params in Google Cloud DNS query during additions conflict resolution

RochesterinNYC:
- GCP API Reference: https://cloud.google.com/dns/api/v1/resourceRecordSets/list#name

This prevents the `get_all` Google DNS HTTP API results from returning all records of the managed zone during resolving additions conflict. We now instead can use the `name` and `type` parameters to filter and only get results/records we're interested in.

**Release note**:

```release-note
Use the Google DNS API `name` and `type` parameters during resolution of 'additions' conflicts to only get records of interest
```


src/gordon_gcp/plugins/service/gdns_publisher.py
```diff
@@ -247,15 +247,16 @@ def _format_resource_record_changes(self, action, resource_record):
...
+            resource_records_url, params=search_params)
         for resp in response:
             for rec in resp['rrsets']:
-                if rec['name'] == record_name and rec['type'] == record_type:
-                    changes['deletions'] = [rec]
-                    break
-
-        return changes
+                changes['deletions'] = [rec]
+                return changes
```
econchick: I haven't looked closely at the failing tests, but there may be cases (i.e. an empty list in `response`) where you won't ever return `changes` from this method.

ocsw: I would also make this an if, not a loop.  if response, changes = response[0]['rrsets'].

RochesterinNYC:
Updated this PR to switch from using `get_all` to `get_json` (since we're returning much less records now).

descent3me:
+1

PR #58, opened 2018-07-17
Increase http logging

descent3me:
This adds more detail to HTTP request and response logging so it is easier to understand what was actually requested.

Fixes #

@spotify/alf 

**Release note**:

```release-note
Increase level of detail in HTTP request/response logging.
```

### DIFF
--- a/src/gordon_gcp/clients/auth.py
+++ b/src/gordon_gcp/clients/auth.py
@@ -53,6 +53,7 @@ import asyncio
 import json
 import logging
 import urllib.parse
+import uuid
 
 import aiohttp
 from google import auth as gauth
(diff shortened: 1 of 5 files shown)


src/gordon_gcp/clients/auth.py
```diff
@@ -173,12 +174,14 @@ def _setup_request_body(self):
...
-        logging.debug(_utils.REQ_LOG_FMT.format(method='POST', url=url))
+        request_id = uuid.uuid4()
+        logging.debug(_utils.REQ_LOG_FMT.format(
+            request_id=request_id, method='POST', url=url, kwargs=None))
         async with self._session.post(url, headers=headers, data=body) as resp:
             log_kw = {
+                'request_id': request_id,
                 'method': 'POST',
-                'url': url,
+                'url': resp.url,
```
ocsw: I'm not sure what the difference is?

ocsw:
+1

src/gordon_gcp/clients/auth.py
descent3me: `url` is the URL that gets passed in, and AFAICT [resp.url](https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientResponse.url) is the [effective URL](https://docs.aiohttp.org/en/stable/client_quickstart.html#passing-parameters-in-urls) that gets called 

I'm not 100% because then I'm sure what [real_url](https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientResponse.real_url) is, but I think `resp.url` is better than just `url` in any case.

PR #60, opened 2018-07-18
 Fix incorrect and superfluous authority logging

descent3me:
Fix incorrect and superfluous authority logging.

Fixes #

@spotify/alf 

**Release note**:

```release-note
Fix incorrect and superfluous authority logging.
```

### DIFF
--- a/src/gordon_gcp/plugins/janitor/authority.py
+++ b/src/gordon_gcp/plugins/janitor/authority.py
@@ -183,9 +183,10 @@ class GCEAuthority:
             try:
                 rrsets.append(self._create_instance_rrset(instance))
             except (KeyError, IndexError) as e:
+                instance_name = instance.get('name')
                 logging.warn(
                     'Could not extract instance information for '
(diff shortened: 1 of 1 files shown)


src/gordon_gcp/plugins/janitor/authority.py
```diff
@@ -141,13 +141,11 @@ def __init__(self, config, crm_client, gce_client, rrset_channel=None,
...
         return sorted(projects - project_blacklist)
 
-    def _filter_results(self, projects, results):
+    def _filter_results(self, results):
         successful_results = []
         for index, result in enumerate(results):
             if isinstance(result, exceptions.GCPHTTPError):
-                project = projects[index]
-                msg = (f'Could not fetch instance list for project {project}: '
-                       f'{result}. Skipping this project.')
```
ocsw: I think we should keep the 'skipping' bit.

descent3me: Hmm, ok

PR #61, opened 2018-07-18
Handle http response errors

descent3me:
This causes an authority to fail if it has an incomplete view of a zone, and changes error handling a bit in order to avoid reimplementing every 4xx/5xx HTTP response code as a specific exception.

Fixes #

@spotify/alf 

**Release note**:

```release-note
* Fail authority if it cannot get a full view of of all instances.
* Simplify HTTP error response handling.
```

### DIFF
--- a/src/gordon_gcp/clients/auth.py
+++ b/src/gordon_gcp/clients/auth.py
@@ -171,7 +171,9 @@ class GAuthClient:
         Raises:
             :exc:`.GCPAuthError`: if no token was found in the
                 response.
-            :exc:`.GCPHTTPError`: if any exception occurred.
+            :exc:`.GCPHTTPError`: if any exception occurred,
+                specifically a :exc:`.GCPHTTPResponseError`, if the
+                exception is associated with a response status code.
         """
         url, headers, body = self._setup_token_request()
         request_id = uuid.uuid4()
@@ -193,7 +195,7 @@ class GAuthClient:
             except aiohttp.ClientResponseError as e:
                 msg = f'[{request_id}] Issue connecting to {resp.url}: {e}'
                 logging.error(msg, exc_info=e)
-                raise exceptions.GCPHTTPError(msg)
+                raise exceptions.GCPHTTPResponseError(msg, resp.status)
 
             response = await resp.json()
             try:
--- a/src/gordon_gcp/exceptions.py
+++ b/src/gordon_gcp/exceptions.py
@@ -19,9 +19,8 @@ from gordon import exceptions as core_exceptions
 
 __all__ = (
     'GCPGordonError', 'GCPGordonJanitorError', 'InvalidMessageError',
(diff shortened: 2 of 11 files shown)


src/gordon_gcp/plugins/service/gdns_publisher.py
```diff
@@ -305,9 +287,13 @@ def _format_resource_record_changes(self, action, resource_record):
         changes_to_publish = self._format_resource_record_changes(
             action, resource_record)
         try:
-            change_id = await self._publish_changes(changes_to_publish,
-                                                    base_changes_url)
-        except exceptions.GCPHTTPConflictError as e:
+            change_id = await self._publish_changes(
```
ocsw: Might want to leave the formatting here as it was, for consistency.

src/gordon_gcp/clients/auth.py
```diff
@@ -193,7 +193,7 @@ def _setup_request_body(self):
             except aiohttp.ClientResponseError as e:
                 msg = f'[{request_id}] Issue connecting to {resp.url}: {e}'
                 logging.error(msg, exc_info=e)
-                raise exceptions.GCPHTTPError(msg)
+                raise exceptions.GCPHTTPResponseError(msg, resp.status)
```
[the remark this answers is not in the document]
descent3me: I thought it was sufficient since a GCPHTTPResponseError is a GCPHTTPError, but being specific is ok, too.

src/gordon_gcp/exceptions.py
```diff
@@ -42,15 +41,14 @@ class InvalidDNSZoneInMessageError(GCPGordonError):
 
 
 class GCPHTTPError(GCPGordonError):
-    """An HTTP error occured."""
+    """An HTTP error occurred."""
```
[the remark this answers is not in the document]
descent3me: I didn't even notice I changed this line...

src/gordon_gcp/plugins/service/gdns_publisher.py
descent3me: I would argue this is more consistent with the surrounding method calls (or at least the two I can see in this file diff)

ocsw: I took another look... They were already inconsistent.  Can you change the rest of the ones in this function too?

            change_id = await self._publish_changes(changes_to_publish,
                                                    base_changes_url)

        change_complete = await self._changes_published(change_id,
                                                        base_changes_url)

descent3me: I think that is outside the scope of this PR.

descent3me:
@ocsw I addressed all comments from your previous review.

src/gordon_gcp/plugins/service/gdns_publisher.py
ocsw: I mean, I would call it leaving things better than you found them, honestly, but I won't hold up the PR because of it.

PR #62, opened 2018-07-19
Add pubsub publisher callback

descent3me:
Improves gpubsub_publisher logging, removing completed futures or (in the future) adding metrics.

Fixes #

@spotify/alf 

**Release note**:

```release-note
Add callback in pubsub publisher.
```

### DIFF
--- a/src/gordon_gcp/plugins/janitor/gpubsub_publisher.py
+++ b/src/gordon_gcp/plugins/janitor/gpubsub_publisher.py
@@ -56,6 +56,7 @@ To use:
 
 import asyncio
 import datetime
+import functools
 import json
 import logging
 import os
@@ -204,6 +205,13 @@ class GPubsubPublisher:
         msg = ('Finished sending reconciliation messages to Google Pub/Sub.')
         logging.info(msg)
 
+    def _message_publish_callback(self, message, future):
+        action = message['action']
+        name = message['resourceRecords']['name']
+        msg_id = future.result()
+        logging.debug(f'Message published for {action}:{name} as {msg_id},'
+                      f'currently tracking {len(self._messages)} messages.')
+
     @threads.threadpool
     def publish(self, message):
         """Publish received change message to Google Pub/Sub.
@@ -215,6 +223,8 @@ class GPubsubPublisher:
         message['timestamp'] = datetime.datetime.utcnow().isoformat()
         bytes_message = bytes(json.dumps(message), encoding='utf-8')
         future = self.publisher.publish(self.topic, bytes_message)
+        future.add_done_callback(
+            functools.partial(self._message_publish_callback, message))
 
         # collect to make sure everything's cleaned up once done
         self._messages.add(future)
--- a/src/gordon_gcp/plugins/janitor/gpubsub_publisher.py
+++ b/src/gordon_gcp/plugins/janitor/gpubsub_publisher.py
@@ -236,6 +236,7 @@ class GPubsubPublisher:
             try:
                 await self.publish(change_message)
             except Exception as e:  # todo
-                logging.error(f'foo: {e}')
+                logging.error('Exception while trying to publish message to '
+                              f' pubsub: {e}')
 
         await self.cleanup()
--- a/src/gordon_gcp/plugins/janitor/gpubsub_publisher.py
+++ b/src/gordon_gcp/plugins/janitor/gpubsub_publisher.py
@@ -209,6 +209,7 @@ class GPubsubPublisher:
         action = message['action']
         name = message['resourceRecords']['name']
         msg_id = future.result()
+        self._messages.remove(future)
         logging.debug(f'Message published for {action}:{name} as {msg_id},'
                       f'currently tracking {len(self._messages)} messages.')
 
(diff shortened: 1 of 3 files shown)


src/gordon_gcp/plugins/janitor/gpubsub_publisher.py
```diff
@@ -223,13 +224,13 @@ def publish(self, message):
...
         future = self.publisher.publish(self.topic, bytes_message)
-        future.add_done_callback(
-            functools.partial(self._message_publish_callback, message))
 
         # collect to make sure everything's cleaned up once done
         self._messages.add(future)
         # TODO (lynn): add metrics.incr/emit call here once aioshumway
         #              is released
+        future.add_done_callback(
+            functools.partial(self._message_publish_callback, message))
```
descent3me: This had to be moved down, in case it completes quickly, it would try to remove itself from the set before it was added.

tests/unit/conftest.py
```diff
@@ -18,6 +18,7 @@
 """
 
 import asyncio
+import concurrent.futures
```
ocsw: Shouldn't this be from concurrent import futures?  Or is that too confusing?

descent3me: I have no strong preference and in this case I probably just copy/pasta'd it from an example on how to test futures. I don't mind more verbose/descriptive unless its a pain in the butt (multiple uses, causing lines to wrap), but I can update if you want.

ocsw: Well, the style guide says 'from', so I guess unless it's too confusing...

descent3me: To be clear the style guide says from is preferred but that both are ok.

ocsw: Er...  I don't think that's the intended interpretation of the guide.  Unless I'm looking in the wrong place or something.

descent3me: I was looking at https://github.com/google/styleguide/blob/gh-pages/pyguide.md#233-decision

ocsw: Huh.  Interesting.  I was looking at 2.2.

src/gordon_gcp/plugins/janitor/gpubsub_publisher.py
ocsw: Ah: 'If the future has already completed or been cancelled, fn will be called immediately.'

tests/unit/plugins/janitor/test_gpubsub_publisher.py
```diff
@@ -101,27 +107,34 @@ def gpubsub_publisher_inst(monkeypatch, kwargs):
...
-    assert 1 == len(gpubsub_publisher_inst._messages)
+    assert expected_num_messages == len(gpubsub_publisher_inst._messages)
 
 
-@pytest.mark.parametrize('raises,exp_log_records', [
-    [False, 1],
-    [Exception('foo'), 2],
+@pytest.mark.parametrize('raises', [
+    False,
+    Exception('foo'),
```
ocsw: This should be True, or else the line below shouldn't have the exception hardcoded.

tests/unit/plugins/janitor/test_gpubsub_publisher.py
```diff
@@ -88,11 +88,17 @@ def gpubsub_publisher_inst(monkeypatch, kwargs):
...
 
 
+@pytest.mark.parametrize('publish_completed', (True, False))
 @pytest.mark.asyncio
-async def test_publish(gpubsub_publisher_inst):
+async def test_publish(gpubsub_publisher_inst, publish_completed, mocker):
     """Publish received messages."""
     datetime.datetime = conftest.MockDatetime
-    msg1 = {'message': 'one'}
+    msg1 = {'action': 'additions', 'resourceRecords': {'name': 'a.b.com.'}}
```
ocsw: I know this isn't strictly part of the PR, but I'd like to see the timestamp and bytes_msg1 lines replaced with hardcoded values - otherwise we're repeating a chunk of the function.

tests/unit/plugins/janitor/test_gpubsub_publisher.py
```diff
@@ -88,11 +88,17 @@ def gpubsub_publisher_inst(monkeypatch, kwargs):
...
+
+    expected_num_messages = 0
+    if not publish_completed:
+        expected_num_messages = 1
+        gpubsub_publisher_inst.publisher.publish.return_value = mocker.Mock()
```
ocsw: Shouldn't this be changing ...publish.return_value.result()?

src/gordon_gcp/plugins/janitor/gpubsub_publisher.py
```diff
@@ -236,6 +247,6 @@ def publish(self, message):
             try:
                 await self.publish(change_message)
             except Exception as e:  # todo
-                logging.error(f'foo: {e}')
+                logging.error(f'Exception while trying to publish record: {e}')
```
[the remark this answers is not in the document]
descent3me: I'll update this message especially because I think 'record' is misleading, but I don't think there will be much confusion in practice since janitor log messages are isolated from core service, and I suspect the error being excepted here will contain Google Pubsub text.

src/gordon_gcp/plugins/janitor/gpubsub_publisher.py
descent3me: Yep.

tests/unit/plugins/janitor/test_gpubsub_publisher.py
```diff
@@ -101,27 +107,34 @@ def gpubsub_publisher_inst(monkeypatch, kwargs):
 
     gpubsub_publisher_inst.publisher.publish.assert_called_once_with(
         gpubsub_publisher_inst.topic, bytes_msg1)
-    assert 1 == len(gpubsub_publisher_inst._messages)
+    assert expected_num_messages == len(gpubsub_publisher_inst._messages)
```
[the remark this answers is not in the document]
ocsw: ok

PR #65, opened 2018-07-31
Drop messages older than a limit

ocsw:
Drop messages older than a configured limit so we don't get overloaded when Gordon hasn't been running for a while.  (And high latency, and high DNS churn...)

Fixes #

**Release note**:

```release-note
Added past_window old-message dropping.
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,17 @@ def install_requires():
     reqs_txt = read('requirements.txt')
     parsed = reqs_txt.split('\n')
     parsed = [r.split('==')[0] for r in parsed]
-    return [r for r in parsed if len(r) > 0]
+    # this breaks the '-e .' in docs-requirements.txt; git lines need to be
+    # added there as well as requirements.txt
+    # (-e . only looks at install_required, not dependency_links, so it won't
+    # pick up the things we're excluding)
+    return [r for r in parsed if len(r) > 0 and not r.startswith('git')]
+
+
+def dependency_links():
+    reqs_txt = read('requirements.txt')
+    parsed = reqs_txt.split('\n')
+    return [r for r in parsed if len(r) > 0 and r.startswith('git')]
 
 
 #####
@@ -112,4 +122,5 @@ setup(
     keywords=KEYWORDS,
     zip_safe=False,
     install_requires=install_requires(),
+    dependency_links=dependency_links()
 )
--- a/src/gordon_gcp/plugins/service/event_consumer.py
+++ b/src/gordon_gcp/plugins/service/event_consumer.py
@@ -143,6 +143,13 @@ class GPSEventConsumerBuilder:
                    'Cloud Pub/Sub is required.')
             errors.append(msg)
 
+        if ('max_msg_age' in self.config and
+                (not isinstance(self.config['max_msg_age'], int) or
+                    self.config['max_msg_age'] < 1)):
+            msg = ('Invalid value for max_msg_age (discard messages older than '
+                   'this many seconds).')
+            errors.append(msg)
+
         if errors:
             exp_msg = 'Invalid configuration:\n'
             for error in errors:
@@ -264,6 +271,7 @@ class GPSEventConsumer:
         self._message_schemas = validator.schemas.keys()
         self._loop = loop
         self._logger = logging.getLogger('')
+        self._max_msg_age = config.get('max_msg_age', 300)
 
     async def handle_message(self, event_msg):
         """Ack Pub/Sub message and update event message history.
@@ -303,6 +311,21 @@ class GPSEventConsumer:
         msg_logger = _utils.GEventMessageLogger(
             self._logger, {'msg_id': msg_id})
         msg_logger.info('Received new message.')
+
+        msg_datetime = pubsub_msg.publish_time
+        # need a non-TZ-aware object in UTC
+        cur_datetime = datetime.datetime.utcnow()
+        msg_age = (cur_datetime - msg_datetime).total_seconds()
+        if msg_age > self._max_msg_age:
+            msg_logger.warn(f'Message is too old ({msg_age} seconds), '
+                            'acking and discarding.')
+            context = {
+                'msg_id': msg_id,
+                'msg_age': msg_age
+            }
+            await self.metrics.incr('msg-too-old', context=context)
+            pubsub_msg.ack()
+            return
 
         try:
             data = json.loads(pubsub_msg.data)
--- a/tests/unit/plugins/service/test_event_consumer.py
+++ b/tests/unit/plugins/service/test_event_consumer.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 import asyncio
+import datetime
 import json
 import threading
 
@@ -51,9 +52,12 @@ def raw_msg_data(creation_audit_log_data):
 
 @pytest.fixture
 def pubsub_msg(mocker, raw_msg_data):
+    mocker.patch(DATETIME_PATCH, conftest.MockDatetime)
     pubsub_msg = mocker.MagicMock(pubsub_v1.subscriber.message.Message)
     pubsub_msg.message_id = 1234
     pubsub_msg.data = bytes(json.dumps(raw_msg_data), encoding='utf-8')
+    # need a non-TZ-aware object in UTC
+    pubsub_msg.publish_time = datetime.datetime.utcnow()
     return pubsub_msg
 
 
@@ -167,7 +171,10 @@ def test_event_consumer_default(mocker, subscriber_client, flow_control,
                                 validator, parser, event_loop, channel_pair,
                                 metrics):
     """GPSEventConsumer implements IEventConsumerClient"""
-    config = {'subscription': '/projects/test-project/subscriptions/test-sub'}
+    config = {
+        'subscription': '/projects/test-project/subscriptions/test-sub',
+        'max_msg_age': 1000
+    }
     success, error = channel_pair
     client = event_consumer.GPSEventConsumer(
         config, success, error, metrics, subscriber_client, flow_control,
@@ -193,7 +200,10 @@ def test_event_consumer_default(mocker, subscriber_client, flow_control,
 @pytest.fixture
 def consumer(subscriber_client, flow_control, validator, event_loop,
              channel_pair, metrics):
-    config = {'subscription': '/projects/test-project/subscriptions/test-sub'}
+    config = {
+        'subscription': '/projects/test-project/subscriptions/test-sub',
+        'max_msg_age': 1000
+    }
     success, error = channel_pair
     parser = parse.MessageParser()
     client = event_consumer.GPSEventConsumer(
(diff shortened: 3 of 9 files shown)


ocsw:
@spotify/alf PTAL
Just need to play with pubsub to be sure it does what I expect.

ocsw:
I used the timestamp on the message itself to avoid having to parse every message even if we're going to reject it.

ocsw:
Ok, apparently the message's publish_time looks like this:
seconds: 1533577139
nanos: 158000000
<class 'google.protobuf.timestamp_pb2.Timestamp'>

Little tweaking coming up.

src/gordon_gcp/plugins/service/event_consumer.py
```diff
@@ -143,6 +144,15 @@ def _validate_config(self):
                    'Cloud Pub/Sub is required.')
             errors.append(msg)
 
+        if 'max_msg_age' not in self.config:
+            self.config['max_msg_age'] = 300
+
+        if (not isinstance(self.config['max_msg_age'], numbers.Number) or
```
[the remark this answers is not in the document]
ocsw: I could.  Not absolutely necessary for it to be an int tho.  Shrug.  I documented it as an int, in fact.  Eh, I'll change it.

src/gordon_gcp/plugins/service/event_consumer.py
```diff
@@ -304,6 +315,15 @@ def _get_and_validate_pubsub_msg_schema(self, pubsub_msg_data):
             self._logger, {'msg_id': msg_id})
         msg_logger.info('Received new message.')
 
+        msg_datetime = pubsub_msg.publish_time
+        # specify UTC as an easy way to get a tz-aware datetime
+        cur_datetime = datetime.datetime.now(datetime.timezone.utc)
+        if (cur_datetime - msg_datetime).total_seconds() > self._max_msg_age:
+            msg_logger.warn('Message is older than max_msg_age, discarding and '
```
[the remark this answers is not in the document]
ocsw: Good call.

ocsw:
@smstone @spotify/alf PTAL

src/gordon_gcp/plugins/service/event_consumer.py
```diff
@@ -143,6 +144,15 @@ def _validate_config(self):
                    'Cloud Pub/Sub is required.')
             errors.append(msg)
 
+        if 'max_msg_age' not in self.config:
+            self.config['max_msg_age'] = 300
+
```
descent3me: Any reason for doing this here?  I wouldn't expect a method called validate to change the state of anything.
I would more expect a default to be set in the consumer init.

src/gordon_gcp/plugins/service/event_consumer.py
```diff
@@ -304,6 +315,21 @@ def _get_and_validate_pubsub_msg_schema(self, pubsub_msg_data):
             self._logger, {'msg_id': msg_id})
         msg_logger.info('Received new message.')
 
+        # The timestamp is in a weird format; see
+        # https://pythonhosted.org/gax-google-logging-v2/google.protobuf.timestamp_pb2.html
+        # It divides a time with nanosecond precision into an integer part and
+        # a nanoseconds part.  To turn them into a regular float, we need to
+        # divide the nanos by a billion (nano = billionth) and add.
+        # E.g.: 1234 seconds and 123456789 nanos -> 1234.123456789
```
descent3me: Could we just do seconds? It would simplify things a bit and I'm not sure we need/have more accuracy than that.

src/gordon_gcp/plugins/service/event_consumer.py
```diff
@@ -304,6 +315,21 @@ def _get_and_validate_pubsub_msg_schema(self, pubsub_msg_data):
...
+        msg_timestamp = (pubsub_msg.publish_time.seconds +
+                         (pubsub_msg.publish_time.nanos / 10**9))
+        msg_age = time.time() - msg_timestamp
+        if msg_age > self._max_msg_age:
+            msg_logger.warn(f'Message is too old ({msg_age} seconds), '
+                            'discarding and acking.')
```
descent3me: Feel free to ignore if this is too picky, but could it be 'acking and discarding'?

src/gordon_gcp/plugins/service/event_consumer.py
```diff
@@ -304,6 +315,21 @@ def _get_and_validate_pubsub_msg_schema(self, pubsub_msg_data):
...
+            pubsub_msg.ack()
+            return
```
descent3me: Could we maybe emit a metric here too?

src/gordon_gcp/plugins/service/event_consumer.py
```diff
@@ -264,6 +274,7 @@ def __init__(self, config, success_channel, error_channel, metrics,
         self._message_schemas = validator.schemas.keys()
         self._loop = loop
         self._logger = logging.getLogger('')
+        self._max_msg_age = config['max_msg_age']
```
descent3me: It might be worth adding `seconds` or even just `_s` in the var names somewhere. It's easy to forget what units things are in.

src/gordon_gcp/plugins/service/event_consumer.py
ocsw: I mean, this is how we've done it elsewhere, but...

src/gordon_gcp/plugins/service/event_consumer.py
ocsw: If we do that we'd have to truncate time.time().  And we could still introduce weird edge cases in  which something past the limit by a fraction of a second doesn't get dropped or vice versa.

src/gordon_gcp/plugins/service/event_consumer.py
ocsw: ok

src/gordon_gcp/plugins/service/event_consumer.py
ocsw: ok

src/gordon_gcp/plugins/service/event_consumer.py
ocsw: Are you including the name in the config file in that?

src/gordon_gcp/plugins/service/event_consumer.py
ocsw: Uh am I losing it, or do we not actually have any metrics in this repo?  The metrics plugin is being passed everywhere but I don't see it being called.

src/gordon_gcp/plugins/service/event_consumer.py
descent3me: Yeah, but my question was more, is that problem? If we miss acking or not acking a message because it arrived within a fraction of second delta from our arbitrarily defined threshold which is likely on the scale of minutes.

ocsw: I mean, it seems like potentially confusing behavior that only saves us like 2 lines of code, vs. doing what I'd consider the "correct" thing.  If everyone hates it i can change it.

src/gordon_gcp/plugins/service/event_consumer.py
ocsw: I would say I'll add a todo and we should have a ticket that covers all of these.

src/gordon_gcp/plugins/service/event_consumer.py
ocsw: Also, this centralizes all the config processing.  Unless you want me to add 'apply_defaults()'.

src/gordon_gcp/plugins/service/event_consumer.py
ocsw: Ok, so I spoke to @descent3me and we're ok with removing the seconds and adding a disclaimer about slight inaccuracy.  BUT.  It turns out that the behavior exhibited by the latest release of google-cloud-pubsub is broken.  The latest code in github has a fix, but evidently they commit everything to master and only release new packages sporadically, so it wasn't clear why the code I get when I download didn't match github...

TLDR: I'm going to pin the library to the commit with the fix and change all this back to using datetimes, which also removes the nanos.

src/gordon_gcp/plugins/service/event_consumer.py
ocsw: Spoke to @descent3me, moving it as requested.

src/gordon_gcp/plugins/service/event_consumer.py
ocsw: Turns out adding the metric is super simple, so I'll just do it.

src/gordon_gcp/plugins/service/event_consumer.py
ocsw: There are other config vars that are seconds but aren't obvious.  We agreed to hold off because they should be consistent, but we shouldn't be doing all of them now.  Also, '300' as the example value should be fairly obvious, one hopes.

ocsw:
Ok, code updates are done.  Just fighting with getting the requirements pin to work.

src/gordon_gcp/plugins/service/event_consumer.py
```diff
@@ -304,6 +312,21 @@ def _get_and_validate_pubsub_msg_schema(self, pubsub_msg_data):
             self._logger, {'msg_id': msg_id})
         msg_logger.info('Received new message.')
 
+        msg_datetime = pubsub_msg.publish_time
+        # need a non-TZ-aware object in UTC
```
descent3me: This comment isn't needed.  I'm all for using UTC everywhere we can, but you should be able to get a timedelta no matter what timezone it has.

tests/unit/plugins/service/test_event_consumer.py
```diff
@@ -351,6 +361,40 @@ def mock_create_gevent_msg(consumer, mocker, monkeypatch):
...
+    pubsub_msg.publish_time -= datetime.timedelta(
+        seconds=(consumer._max_msg_age + 1))
+    context = {
+        'msg_id': pubsub_msg.message_id,
+        'msg_age': consumer._max_msg_age + 1
+    }
+
+    event_msg = event_consumer.GEventMessage(
+        pubsub_msg, creation_audit_log_data, phase='consume')
+    mock_create_gevent_msg.return_value = event_msg
```
descent3me: Why are you mocking this if we're also asserting it's not called?

tests/unit/plugins/service/test_event_consumer.py
```diff
@@ -351,6 +361,40 @@ def mock_create_gevent_msg(consumer, mocker, monkeypatch):
...
+
+    mock_run_coro_threadsafe = mocker.Mock()
+    patch = ('gordon_gcp.plugins.service.event_consumer.asyncio.'
+             'run_coroutine_threadsafe')
+    monkeypatch.setattr(patch, mock_run_coro_threadsafe)
```
descent3me: We could skip this if we just verify there was nothing put on the success_channel instead.

tests/unit/plugins/service/test_event_consumer.py
ocsw: Well, a mock is needed to make the assertion, but I copypastad that.  I'll take another look.

tests/unit/plugins/service/test_event_consumer.py
ocsw: ok

src/gordon_gcp/plugins/service/event_consumer.py
ocsw: datetimes work in strange ways.  There are aware and non-aware ones.  The non-aware ones have to be created with a particular TZ, but then don't record the TZ.  They're giving us a non-aware UTC datetime, so that's what we need to compare with.

To clarify:
```
>>> import datetime
>>> datetime.datetime.now()
datetime.datetime(2018, 8, 7, 12, 31, 48, 589097)
>>> datetime.datetime.utcnow()
datetime.datetime(2018, 8, 7, 16, 31, 54, 83595)
>>> datetime.datetime.now(datetime.timezone.utc)
datetime.datetime(2018, 8, 7, 16, 32, 7, 868219, tzinfo=datetime.timezone.utc)
```

The first two are non-aware objects based on two different TZs.  The third is TZ-aware and therefore can be compared to TZ-aware objects in any TZ.

ocsw: (The docs call non-aware "naive" btw.)

ocsw: (And they won't let you operate on aware and naive objects in the same expression.)

tests/unit/plugins/service/test_event_consumer.py
ocsw: (also copypasta)

src/gordon_gcp/plugins/service/event_consumer.py
descent3me: Thanks, I've never used datetime with timezones before, so I wasn't aware of this limitation. Using timezones should be the exception rather than without.

setup.py
```diff
@@ -56,7 +56,13 @@ def install_requires():
     reqs_txt = read('requirements.txt')
     parsed = reqs_txt.split('\n')
     parsed = [r.split('==')[0] for r in parsed]
-    return [r for r in parsed if len(r) > 0]
+    return [r for r in parsed if len(r) > 0 and not r.startswith('git')]
```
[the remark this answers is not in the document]
ocsw: (-e in requirements.txt, that is, not in docs-requirements.txt)

ocsw:
@descent3me PTAL

ocsw:
@smstone PTAL

PR #66, opened 2018-08-01
Default max results to 100 when listing instances

smstone:
This changes the default maxResults to 100 when listing instances. We've found we occasionally receive HTTP 500s from Google APIs with projects with a lot of instance churn when using the max results of 500.

Fixes #

**Release note**:

```release-note

```

@spotify/alf


ocsw:
+1

PR #67, opened 2018-08-01
Add 403 to project skip response codes

smstone:
We may see a HTTP 403 returned when fetching projects that are either new or testing.

Fixes #

**Release note**:

```release-note

```

@spotify/alf


descent3me:
It also happens when they are deleted (or at least that's what it looks like when I try to load the project manually in the console).  I don't know if it's because they were recently deleted or deleted in some different way, or if the error message just makes it look like they were deleted because I as a user am unauthorized to see if it exists or not.

+1

PR #69, opened 2018-08-01
Fix rrset paging in dns client

descent3me:
We were previously stripping the returned fields down to a specific set, which did not contain the nextPageToken field for paging.

Fix rrset paging in dns client. We were previously stripping the returned fields down to a specific set, which did not contain the nextPageToken field for paging.

Fixes #

@spotify/alf 

**Release note**:

```release-note
Fix rrset paging in dns client.
```


ocsw:
+1

PR #70, opened 2018-08-01
Convert DNS zone to managed zone before querying GDNS api

smstone:
`get_records_for_zone` expects a managed zone format for the zone used to query the GDNS api.

Fixes #

**Release note**:

```release-note

```
@spotify/alf


ocsw:
We can't do it this way in the public version.  We were going with the assumption that the managed_zone could be anything.  It was supposed to have been a config setting tho.

PR #73, opened 2018-08-06
Use automated zone naming

descent3me:
Do DNS zone to Google managedZone identifier in only one place, GDNSClient, which also does an automated conversion rather than allowing any arbitrary string.

Fixes #

@spotify/alf 

**Release note**:

```release-note
* Remove managed zone from plugin configuration in favor of automated conversion from DNS zone.
* Do DNS zone to managed zone conversion only in GDNSClient.
```


.gitignore
```diff
@@ -44,6 +44,9 @@ coverage.xml
 .project
 .pydevproject
 
+# Vim
+*.swp
```
ocsw: Maybe we should add this everywhere?  Or alternatively, you can add it to your .gitignore_global.  That’s what I did with .vscode:

[core]
    pager = less -F
    excludesfile = /Users/zephyr/.gitignore_global
$ cat ~/.gitignore_global
.python-version
.vscode

Then we don't have to have everyone's editors in every repo.

ocsw:
You'll also need to remove it from gordon.toml.example.

descent3me:
D'oh! Thanks, I was waiting until the last commit to do it, then I forgot to do it when I finally got there.

.gitignore
descent3me: I took the fact that several things we don't use are already there as indication it would be okay to expand on, but I'm fine removing it if this policy we don't want to set a precedent for.

ocsw: I'm open to it but only if we discuss it as a group first.  I'd like to have a clear global policy.

src/gordon_gcp/clients/gdns.py
```diff
@@ -110,15 +110,41 @@ def _parse_resp_to_records(self, response, records):
...
+        Returns:
+            str of managed zone name.
+
+        """
+        if zone.endswith('.in-addr.arpa.'):
+            return 'reverse-' + '-'.join(zone.split('.')[1:3])
+        return '-'.join(zone.split('.')[:-1])
+
     async def get_records_for_zone(self, zone):
         """Get all resource record sets for a particular managed zone.
```
ocsw: This should probably be 'managed zone'.

tests/unit/clients/test_gdns.py
```diff
@@ -101,10 +91,18 @@ def client(mocker):
...
         assert all(
             [isinstance(r, gdns.GCPResourceRecordSet) for r in records])
         assert 6 == len(records)
 
     assert 1 == len(caplog.records)
+
+
+@pytest.mark.parametrize('dns_zone,exp_managed_zone', [
+    ('example.com.', 'example-com'),
+    ('30.20.10.in-addr.arpa.', 'reverse-20-10')
```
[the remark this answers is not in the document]
descent3me: Would you prefer the client raise an error, or what do you expect should happen in that case? 
I like that you brought it up, because I hadn't thought about it but I also like the idea of keeping the client as dumb as possible, and allow errors to pass through and leave the policing of what is or is not a valid zone to the plugins. In that case no error would be raised, but I wouldn't necessarily want to give an example of using the client that way by adding a test for it.

src/gordon_gcp/clients/gdns.py
descent3me: Do you mean "*not* be managed zone"?  I was on the fence, but I can remove it, if you like.

src/gordon_gcp/plugins/janitor/reconciler.py
```diff
@@ -282,11 +282,10 @@ def _remove_soa_and_root_ns(self, zone, rrsets):
         Returns:
             tuple[list[rrset], list[rrset]]: The missing and extra rrset lists.
         """
-        desired_rrsets = [
-            gdns.GCPResourceRecordSet(**record) for record in rrsets
-        ]
+        desired_rrsets = gdns.GDNSClient.get_rrsets_as_objects(rrsets)
```
smstone: just wondering, why did you chose to add `get_rrsets_as_objects` as a static method then call it as such here? this class already has the `dns_client` available, so could just be `self.dns_client.get_rrsets_as_objects(rrsets)`? apologies if i'm missing something

src/gordon_gcp/clients/gdns.py
smstone: i think @ocsw was referring to the function name?

src/gordon_gcp/plugins/service/gdns_publisher.py
```diff
@@ -233,78 +204,60 @@ def _format_resource_record_changes(self, action, resource_record):
...
-            changes_to_publish = await self._handle_additions_conflict(
-                changes_to_publish, resource_records_url)
-            change_id = await self._publish_changes(changes_to_publish,
-                                                    base_changes_url)
-
-        change_complete = await self._changes_published(change_id,
-                                                        base_changes_url)
-        if not change_complete:
+
+            deletions = await self._get_conflicting_rrsets(zone, changes)
```
smstone: is this accurate that it's only deletions? https://github.com/spotify/gordon-gcp/pull/73/commits/84dea31582fb68488230f2c7d2eb5c03880faa30#diff-8f53676280c6a88b173ebd83ba0803c0R199 says dict containing original additions plus deletions

tests/unit/plugins/service/test_gdns_publisher.py
```diff
@@ -308,56 +275,43 @@ def test_implements_interface(mocker, config, metrics):
...
 async def test_handle_message_handles_update_conflict(
         gdns_publisher_instance, event_message,
-        matching_zone_records, rrsets_url, changes_url,
         initial_changes_req, handled_conflict_changes_req,
         initial_changes_pending_json_resp, initial_changes_resp, caplog,
-        mocker):
+        mocker, mock_dns_client):
     """Ensure changes with update conflicts are successfully published"""
-    expected_change_id = json.loads(initial_changes_pending_json_resp)['id']
+    # expected_change_id = json.loads(initial_changes_pending_json_resp)['id']
```
smstone: can this be removed?

tests/unit/clients/test_gdns.py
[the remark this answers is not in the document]
ocsw: Yeah I guess we should leave it.

src/gordon_gcp/clients/gdns.py
```diff
@@ -97,47 +97,109 @@ class GDNSClient(http.AIOConnection):
...
+            str of managed zone name.
+
+        """
+        if zone.endswith('.in-addr.arpa.'):
+            return 'reverse-' + '-'.join(zone.split('.')[1:3])
+        return '-'.join(zone.split('.')[:-1])
 
-    async def get_records_for_zone(self, zone):
+    async def get_records_for_zone(self, zone, params=None):
         """Get all resource record sets for a particular managed zone.
```
ocsw: Actually I meant that you're actually passing in a DNS zone, so the docstring (and probably the function name) should reflect that.

src/gordon_gcp/clients/gdns.py
```diff
@@ -110,15 +110,41 @@ def _parse_resp_to_records(self, response, records):
             rrset = GCPResourceRecordSet(**record)
             records.append(rrset)
 
+    def get_managed_zone(self, zone):
+        """Get the GDNS managed zone name for a DNS zone.
+
+        Google uses custom string names with specific `requirements
```
ocsw: This should be added in the top-level docs somewhere.

src/gordon_gcp/plugins/janitor/reconciler.py
ocsw: So my usual MO was to make methods that didn't depend on instance variables static.  It indicates that separation.  FWIW, tho, the new style guide hates static methods and tells you to make either instance-level methods or module-level functions.

src/gordon_gcp/plugins/service/enricher.py
```diff
@@ -222,16 +216,10 @@ def _get_internal_ip(self, instance_data):
...
         fqdn = self._get_fqdn(instance_name)
 
-        # Get the Google DNS A records for managed zone with name matching the
-        # instance fqdn
-        project = self.config['project']
-        managed_zone = self.config['managed_zone']
-        resource_records_url = GCEEnricher.RESOURCE_RECORDS_ENDPOINT.format(
-            project=project, managedZone=managed_zone)
-
+        # Get the A records with name matching the instance FQDN
```
ocsw: FWIW, 'records in the managed zone with' isn't redundant; for private zones you can have the same name/type in two zones.  (I think?)

tests/unit/clients/test_gdns.py
```diff
@@ -86,26 +86,31 @@ def test_get_rrsets_as_objects(rrset_dict):
...
 @pytest.mark.asyncio
+@pytest.mark.parametrize('params', [
+    None,
+    {'fields': 'some-other-field,another-field'},
+    {'name': 'example.com.', 'type': 'A'}
+])
 async def test_get_records_for_zone(fake_response_data, client, caplog,
-                                    monkeypatch):
-    mock_get_json_called = 0
+                                    mocker, params):
```
ocsw: mocker doesn't seem to be used

tests/unit/clients/test_gdns.py
```diff
@@ -86,26 +86,31 @@ def test_get_rrsets_as_objects(rrset_dict):
...
-    monkeypatch.setattr(client, 'get_json', mock_get_json)
+    client.get_json = mock_get_json
 
     url = f'{client._base_url}/managedZones/a-zone/rrsets'
     with aioresponses() as mocked:
         mocked.get(url, status=200)
         # paginated requests
         mocked.get(url, status=200)
-        records = await client.get_records_for_zone('a.zone.')
+        records = await client.get_records_for_zone('a.zone.', params=params)
```
ocsw: Should we be testing that await self.get_json(url, params=params) was called with the right params?

src/gordon_gcp/clients/gdns.py
descent3me: Well, it's a little trickier than that. The records for the relevant DNS zone is requested (which is why that is indicated in args part of the docstring), but what is returned is the records for the managed zone that corresponds to that DNS zone (which is why I left it in this part of the docstring).  Because gordon-gcp currently puts all records in the same DNS zone, this is not an important distinction, *but* if any user of the client were to pass in reverse records it is significant, since the scheme we use for now has a many to one relationship of DNS zone[s] -> managed zone because we reduce to the two most significant octets in reverse zones.  For example `get_records_for_zone('30.20.10.in-addr.arpa')` returns records from that `30.20.10.in-addr.arpa`, but also might return records from `31.20.10.in-addr.arpa`.

src/gordon_gcp/clients/gdns.py
```diff
@@ -153,11 +153,12 @@ def get_managed_zone(self, zone):
 
         # to limit the amount of data across the wire; also makes it
         # easier to create GCPResourceRecordSet instances
-        fields = ('rrsets/name,rrsets/kind,rrsets/rrdatas,'
-                  'rrsets/type,rrsets/ttl,nextPageToken')
-        params = {
-            'fields': fields,
-        }
+        if not params:
```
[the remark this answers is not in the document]
descent3me: Ah,  good point.  I'll bump it down into the if block where fields are added and nix the bit about data, because it seems like it only removes one field, and that seems less relevant

src/gordon_gcp/plugins/janitor/reconciler.py
descent3me: Yeah, it was basically what ocsw said, but I don't mind doing it as an instance method if you folks prefer. It makes it a little more flexible if we could ever imagine some config info being related to the process, in the same way I left `get_managed_zone` as an instance method, because I thought it was plausible want to drop the auto-translation and instead pass in a config when initializing the dns client that contains a map of dns->managed zones at some point.

src/gordon_gcp/plugins/service/enricher.py
descent3me: I can't see why you couldn't, but I don't know if that's distinction necessary to make here, since that would change architecture assumptions in a lot of places in gordon.

src/gordon_gcp/plugins/service/gdns_publisher.py
descent3me: I changed the API of the method a bit. Instead of passing in a changes dict, and adding a deletions to it inside the function, we get just the deletions back here, and add them to the changes dict on the line below. 

It made it easier for me to understand what was going on, but it's not super important to me, if it makes it more confusing for others.

tests/unit/clients/test_gdns.py
descent3me: Hmm, nope, wonder what was going on there...  Good catch!

tests/unit/plugins/service/test_gdns_publisher.py
descent3me: Yep, thanks!

src/gordon_gcp/plugins/janitor/reconciler.py
ocsw: I'm ok either way.

src/gordon_gcp/clients/gdns.py
ocsw: I would argue that the DNS zone in that case is 20.10.in-addr.arpa.  Just depends where you stick the SOA and NS.  I don't care super much as long as it's clear that it's taking one thing and returning (essentially) another.

src/gordon_gcp/plugins/service/enricher.py
ocsw: I did bear it in mind in places, but it's probably an acceptable simplification for most use cases.

src/gordon_gcp/clients/gdns.py
```diff
@@ -97,6 +97,7 @@ class GDNSClient(http.AIOConnection):
             object attached to :obj:`auth_client` if not provided.
     """
     BASE_URL = 'https://www.googleapis.com/dns'
+    DNS_CHANGES_DONE = 'done'
```
ocsw: # see https://cloud.google.com/dns/api/v1/changes#resource

src/gordon_gcp/plugins/service/gdns_publisher.py
```diff
@@ -195,24 +183,7 @@ def _format_resource_record_changes(self, action, resource_record):
...
-        """
-        resp = await self.http_client.request(
-            'post', base_changes_url, json=changes)
-        resp_dict = json.loads(resp)
-
-        # TODO: create another task to measure propagation time
-        return resp_dict['id']
-
-    async def _handle_additions_conflict(self, changes, resource_records_url):
+    async def _get_conflicting_rrsets(self, zone, changes):
```
ocsw: The docstring is now super incorrect.

src/gordon_gcp/plugins/service/gdns_publisher.py
```diff
@@ -233,78 +204,60 @@ def _format_resource_record_changes(self, action, resource_record):
...
-        response = await self.http_client.get_json(
-            resource_records_url, params=search_params)
-        if response['rrsets']:
-            changes['deletions'] = response['rrsets']
-        return changes
+        return await self.dns_client.get_records_for_zone(
+            zone, params=search_params)
 
-    async def _changes_published(self, change_id, base_changes_url):
+    async def _is_change_done(self, zone, change_id):
```
ocsw: We should put something in the name or docstring to note that there is a timeout applied.

src/gordon_gcp/plugins/service/gdns_publisher.py
```diff
@@ -233,78 +204,60 @@ def _format_resource_record_changes(self, action, resource_record):
...
-            resp = await self.http_client.get_json(change_id_url)
-            if resp['status'] == GDNSPublisher.DNS_CHANGES_DONE:
+            if await self.dns_client.is_change_done(zone):
                 return True
             await asyncio.sleep(1)
 
         return False
 
-    async def _dispatch_changes(self, resource_record, managed_zone, action,
+    async def _dispatch_changes(self, resource_record, zone, action,
```
ocsw: I would prefer to call this dns_zone.

src/gordon_gcp/plugins/service/gdns_publisher.py
ocsw: That docstring is wrong now.

tests/unit/plugins/service/test_gdns_publisher.py
```diff
@@ -280,15 +249,13 @@ def mock_sleep(mocker):
 @pytest.mark.filterwarnings(
     "ignore:coroutine 'test_gpsthread_add_task.<locals>.noop' was never "
```
ocsw: Try removing this and seeing if it's still broken. :)

tests/unit/plugins/service/test_gdns_publisher.py
```diff
@@ -308,56 +275,43 @@ def test_implements_interface(mocker, config, metrics):
...
-        initial_changes_req, initial_changes_resp,
-        initial_changes_pending_json_resp, mock_sleep, caplog):
+        gdns_publisher_instance, event_message, mock_dns_client,
+        initial_changes_req, mock_sleep, caplog, mocker):
     """Ensure exception raised when publish wait timeout exceeded."""
-    gdns_publisher_instance.publish_wait_timeout = 0.0001
+    start = datetime.datetime(2018, 1, 1, 11, 30, 0)
+    mockdatetime = mocker.Mock()
+    mockdatetime.now = mocker.Mock(side_effect=[
+        start, start, start + datetime.timedelta(seconds=100)])
```
ocsw: Would be nice to replace 100 with gdns_publisher_instance.config['publish_wait_timeout'] + 1.

tests/unit/plugins/service/test_gdns_publisher.py
```diff
@@ -371,34 +325,19 @@ def test_implements_interface(mocker, config, metrics):
 
 @pytest.mark.parametrize('exception_args,http_exception', http_exceptions)
 @pytest.mark.asyncio
-async def test__dispatch_changes_http_exceptions_raised(
+async def test_dispatch_changes_http_exceptions_raised(
```
ocsw: The second _ was deliberate.  The first one is a word separator, the second is the leading _ from the method.

ocsw:
handled_conflict_changes_done is no longer used

ocsw:
@spotify/alf I've made all the changes that were suggested on here, and added some docs.  Just have one test I'm debating adding.  Anyone want to do one final sanity check?

ocsw:
Added one more test, fixed a super subtle bug...

docs/config-service.rst
```diff
@@ -70,22 +76,16 @@ All configuration options above in the general ``[gcp]`` may be used here. Addit
...
 ``[gcp.publisher]``
 ~~~~~~~~~~~~~~~~~~~
 
-All configuration options above in the general ``[gcp]`` may be used here. Additional Google Cloud DNS configuration options are:
-
-.. describe:: dns_zone="STR"
-
-    `Required`: DNS zone to validate the correctness of A records before publishing. Must be a fully-qualified domain name (FQDN), ending in ``.``, e.g. ``example.com.``.
+All configuration options above in the general ``[gcp]`` may be used here.  If
+``dns_zone`` isn't present here, it must be in ``[gcp]``.Additional Google
```
smstone: missing a space before Additional

ocsw: I don't know why this was already hidden, but I fixed it.

ocsw:
@smstone added a couple of reference links

src/gordon_gcp/clients/gdns.py
```diff
@@ -97,47 +98,116 @@ class GDNSClient(http.AIOConnection):
...
-        fields = ('rrsets/name,rrsets/kind,rrsets/rrdatas,'
-                  'rrsets/type,rrsets/ttl,nextPageToken')
-        params = {
-            'fields': fields,
-        }
+        if zone.endswith('.in-addr.arpa.'):
+            return self.REVERSE_PREFIX + '-'.join(zone.split('.')[-5:-3])
+        return '-'.join(zone.split('.')[:-1])
+
+    async def get_records_for_zone(self, dns_zone, params=None):
```
econchick: Can we get a docstring for the new `params` arg?

ocsw: added

src/gordon_gcp/clients/gdns.py
```diff
@@ -97,47 +98,117 @@ class GDNSClient(http.AIOConnection):
...
+        url = f'{self._base_url}/managedZones/{managed_zone}/rrsets'
+
+        if not params:
+            params = {}
+        else:
+            # otherwise we can break things in callers because we're actually
+            # modifying the same object; see
+            # https://en.wikipedia.org/wiki/Evaluation_strategy#Call_by_sharing
+            # https://jeffknupp.com/blog/2012/11/13/is-python-callbyvalue-or-callbyreference-neither/
+            params = copy.deepcopy(params)
```
econchick: Hmm, I'm having trouble understanding what the problem is that creating a new dictionary object is trying to solve.

What it _seems_ like this is trying to solve is the issue with `params` mutating. For instance (walking thru this mostly for my benefit to make sure I understand):

```pycon
>>> some_params = {'foo': 'bar'}
>>> def foo(params=None):
...   if not params:
...     params = {}
...   print(f'1: {params}')
...   params['a_new_key'] = 'some value'
...   print(f'2: {params}')
...

# no `params` arg given, `params` value is not mutated
>>> foo()
1: {}
2: {'a_new_key': 'some value'}
>>> foo()
1: {}
2: {'a_new_key': 'some value'}

# `params` arg given, `params` value is mutated
>>> foo(some_params)
1: {'foo': 'bar'}
2: {'foo': 'bar', 'a_new_key': 'some value'}
>>> foo(some_params)
# potentially expecting {'foo': 'bar'} here
1: {'foo': 'bar', 'a_new_key': 'some value'} 
2: {'foo': 'bar', 'a_new_key': 'some value'}
```

So then creating a new object via deep copy fixes that:

```pycon
>>> some_params = {'foo': 'bar'}
>>> def foo(params=None):
...   if not params:
...     params = {}
...   else:
...     params = copy.deepcopy(params)
...   print(f'1: {params}')
...   params['a_new_key'] = 'some value'
...   print(f'2: {params}')
...
>>> foo(some_params)
1: {'foo': 'bar'}
2: {'foo': 'bar', 'a_new_key': 'some value'}
>>> foo(some_params)
1: {'foo': 'bar'}
2: {'foo': 'bar', 'a_new_key': 'some value'}
```

But I don't know if that's a problem in our case, as when we call this coroutine method `get_records_for_zone` in the enricher, we don't re-use `params`; we always create a new object, `search_params`, each time we call the method. A simplified calling path:

```pycon
# simplification of this current method 
>>> def get_records_for_zone(params=None):
...   if not params:
...     params = {}
...   print(f'1: {params}')
...   if not 'a_new_key' in params:
...     params['a_new_key'] = 'some_value'
...   print(f'2: {params}')

# simplification of the calling method that uses the params arg
>>> def _get_matching_records():
...   search_params = {'name': 'foo.spotify.net', 'type': 'A'}
...   get_records_for_zone(search_params)

# _get_matching_records called per message received
>>> _get_matching_records()
1: {'name': 'foo.spotify.net', 'type': 'A'}
2: {'name': 'foo.spotify.net', 'type': 'A', 'a_new_key': 'some_value'}
>>> _get_matching_records()
1: {'name': 'foo.spotify.net', 'type': 'A'}
2: {'name': 'foo.spotify.net', 'type': 'A', 'a_new_key': 'some_value'}
```

*But* I recognize that I've been without Gordon context for the past 3 weeks, so I could very well be missing something. It's also friday :D

ocsw: Those top two code blocks, yes.  The problem isn't in the call from the enricher, it's from the tests.

src/gordon_gcp/plugins/service/gdns_publisher.py
```diff
@@ -195,116 +183,76 @@ def _format_resource_record_changes(self, action, resource_record):
...
 
         Args:
+            zone (str): The DNS zone name.
             changes (dict): Changes containing an addition conflict.
-            resource_records_url (str): The URL to use to get the zone's
-                resource records.
 
         Returns:
-            dict containing original additions plus deletions.
+            list of rrsets: The GDNS version of the records with the conflict.
```
smstone: what does `GDNS version` mean here?

ocsw: There's the version we're trying to add, and the version GDNS has.  Version meaning same keys, different values.  I'm trying to come up with a better way to put it, but having difficulty.

ocsw: 'get the GDNS data for records with a conflict'?

src/gordon_gcp/clients/gdns.py
```diff
@@ -97,47 +97,111 @@ class GDNSClient(http.AIOConnection):
...
 
-    async def get_records_for_zone(self, zone):
-        """Get all resource record sets for a particular managed zone.
+    async def get_records_for_zone(self, dns_zone, params=None):
+        """Get all resource record sets for a managed zone, using the DNS zone.
 
         Args:
-            zone (str): Desired managed zone to query.
+            dns_zone (str): Desired DNS zone to query.
+            params (dict): Optional set of search parameters to pass to GDNS.
```
econchick: I appreciate a little more clarity: these aren't search parameters (it's just the calling method refers to them as such), they are query parameters. Perhaps something like 

```
params (dict): (optional) Additional query parameters for HTTP requests to the GDNS API.
```

(keeping in line with how we typically [denote optional params](https://github.com/spotify/gordon-gcp/search?q=optional&unscoped_q=optional))

ocsw: ok

econchick:
@ocsw mind cleaning up that `temp` commit?

src/gordon_gcp/plugins/service/gdns_publisher.py
ocsw: """Get conflicting records as reflected in GDNS.
            (list[rrset]): The conflicting records as reflected in GDNS.

How's that?

ocsw: Ok, I refactored.  @econchick ?

docs/config-janitor.rst
```diff
@@ -74,9 +74,11 @@ All configuration options from the general ``[gcp]`` section may be used here.
 
 Additional plugin-specific configuration is needed:
 
-.. option:: zone="STR"
+.. option:: dns_zone="STR"
 
-    `Required`: ID of a Cloud DNS managed zone that instance resource records should belong to.
+    `Required`: DNS zone to validate the correctness of A records before publishing. Must be a fully-qualified domain name (FQDN), ending in ``.``, e.g. ``example.com.``.  If it's a reverse zone, it must be in the form 'A.B.in-addr.arpa.'.  This setting must be either in this section, or in both the ``[gcp.event_consumer]`` and ``[gcp.publisher]`` sections.
```
descent3me: Is "validate the correctness" really the correct wording for how it's used everywhere?  Either way, the section titles are incorrect for janitor.

ocsw: Good catch.

ocsw: Fixed.

src/gordon_gcp/plugins/service/gdns_publisher.py
econchick: :+1:

PR #74, opened 2018-08-14
Add metrics to janitor plugins

smstone:
Adds metrics around janitor GCP plugins

Fixes #

**Release note**:

```release-note
Add metrics to janitor GCP plugins
```

@spotify/alf

### DIFF
--- a/src/gordon_gcp/plugins/janitor/authority.py
+++ b/src/gordon_gcp/plugins/janitor/authority.py
@@ -61,11 +61,13 @@ class GCEAuthorityBuilder:
 
     Args:
         config (dict): plugin-specific configuration.
+        metrics (obj): :interface:`IMetricRelay` implementation.
         rrset_channel (asyncio.Queue): channel to send resource record messages
             to.
     """
-    def __init__(self, config, rrset_channel, **kwargs):
+    def __init__(self, config, metrics, rrset_channel, **kwargs):
         self.config = config
+        self.metrics = metrics
         self.rrset_channel = rrset_channel
         self.kwargs = kwargs
         self.session = None
@@ -107,7 +109,7 @@ class GCEAuthorityBuilder:
         crm_client = self._get_crm_client(keyfile_path, scopes)
         gce_client = self._get_gce_client(keyfile_path, scopes)
 
-        return GCEAuthority(self.config, crm_client, gce_client,
+        return GCEAuthority(self.config, self.metrics, crm_client, gce_client,
                             self.rrset_channel, **self.kwargs)
 
 
@@ -117,6 +119,7 @@ class GCEAuthority:
 
     Args:
         config (dict): plugin-specific configuration.
+        metrics (obj): :interface:`IMetricRelay` implementation.
         crm_client (.GCRMClient): client used to fetch GCE projects.
         gce_client (.GCEClient): client used to fetch instances for a project.
         rrset_channel (asyncio.Queue): channel to send resource record messages
@@ -125,9 +128,10 @@ class GCEAuthority:
 
     PROJECT_SKIP_RESP_CODES = {403, 404, 410}
 
-    def __init__(self, config, crm_client, gce_client, rrset_channel=None,
-                 **kwargs):
+    def __init__(self, config, metrics, crm_client, gce_client,
+                 rrset_channel=None, **kwargs):
         self.config = config
+        self.metrics = metrics
         self.crm_client = crm_client
         self.gce_client = gce_client
         self.rrset_channel = rrset_channel
@@ -139,9 +143,11 @@ class GCEAuthority:
     async def _get_projects(self):
         projects = await self._get_active_project_ids()
         project_blacklist = set(self.config.get('project_blacklist', []))
-        # TODO: emit a metric for all projects and a metric for
-        # project - blacklist.
-        return sorted(projects - project_blacklist)
+        sorted_projects = sorted(projects - project_blacklist)
+        context = {'plugin': 'gceauthority'}
+        await self.metrics.set(
+            'projects', len(sorted_projects), context=context)
+        return sorted_projects
 
     def _filter_results(self, results):
         successful_results = []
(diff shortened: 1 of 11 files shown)


tests/unit/plugins/janitor/test_authority.py
```diff
@@ -139,12 +141,28 @@ def _create_fake(klass, *args, **kwargs):
     assert expected_msg == (await rrset_channel.get())
     # run also calls self.cleanup at the end
     assert (await rrset_channel.get()) is None
+    gce_authority.metrics.timer.assert_called_once_with(
+        'janitor-authority-runtime')
+    gce_authority.metrics.timer(None).start.assert_called_once_with()
```
descent3me: This works because we've mocked timer to return the same mock object every time?  I'm having a little trouble finding the metrics definition to be sure.

src/gordon_gcp/plugins/janitor/authority.py
```diff
@@ -202,18 +210,27 @@ def _create_msgs(self, instances):
...
+            total_rrsets = len(rrsets)
+            msg = f'[{zone}] Found {total_rrsets} rrsets for zone in GCE.'
             logging.info(msg)
+            context = {
+                'rrsets': total_rrsets,
+                'zone': zone
+            }
+            await self.metrics.incr('janitor-authority-rrsets',
+                                    context=context)
             await self.rrset_channel.put(rrset_msg)
```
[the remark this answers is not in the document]
smstone: @descent3me thanks - i'll switch it to `set` and i think your other idea is good too.

src/gordon_gcp/plugins/janitor/authority.py
```diff
@@ -202,18 +210,27 @@ def _create_msgs(self, instances):
     async def run(self):
         """Batch instance data and send it to the :obj:`self.rrset_channel`.
         """
+        timer = self.metrics.timer('janitor-authority-runtime')
+        timer.start()
```
econchick: IIRC this is a coroutine, so it should be `await timer.start()` (and same as `await timer.stop()`)

smstone:
@econchick i've updated according to @descent3me comments. can you PTAL?

tests/unit/conftest.py
```diff
@@ -282,20 +282,38 @@ def emulator(monkeypatch):
...
-        incr(*args, **kwargs)
 
-    async def set_func(*args, **kwargs):
-        set_(*args, **kwargs)
+class MetricRelayStub:
+    def __init__(self, mocker, config):
+        self.config = config
+        self.incr_mock = mocker.Mock()
+        self.set_mock = mocker.Mock()
+        self.timer_mock = mocker.Mock()
```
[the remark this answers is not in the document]
econchick: @smstone what do you think of @descent3me comment? It makes sense to have it semi-private, particularly if we [address this](https://github.com/spotify/gordon-gcp/pull/74#discussion_r211009288).

src/gordon_gcp/plugins/janitor/authority.py
```diff
@@ -139,9 +143,13 @@ def __init__(self, config, crm_client, gce_client, rrset_channel=None,
...
         project_blacklist = set(self.config.get('project_blacklist', []))
-        # TODO: emit a metric for all projects and a metric for
-        # project - blacklist.
-        return sorted(projects - project_blacklist)
+        sorted_projects = sorted(projects - project_blacklist)
+        context = {
+                'plugin': 'gceauthority'
+            }
+        await self.metrics.set(
+            'janitor-projects', len(sorted_projects), context=context)
```
econchick: What do you think of making use of metrics config ([example](https://github.com/spotify/gordon/blob/develop/gordon/metrics/ffwd.py#L29)), and setting the key to be `gordon-janitor`, so then we don't have to prefix the metric names with `janitor-` everywhere. 

Also, since this is GCP-specific, if/when we have another provider(s), we may want to make it more obvious where it makes sense, e.g. `gcp-projects`, `gcp-plugin-runtime`, etc.

src/gordon_gcp/plugins/janitor/gpubsub_publisher.py
```diff
@@ -238,16 +238,21 @@ def publish(self, message):
...
+        context = {'plugin': 'gpubsub-publisher'}
+        timer = self.metrics.timer('janitor-plugin-runtime', context=context)
+        await timer.start()
         while True:
             change_message = await self.changes_channel.get()
             if change_message is None:
                 break
-            # TODO (lynn): emit metric of message received once
-            #              aioshumway is released
+            await self.metrics.incr('janitor-msg-recv', context=context)
```
econchick: `msg-recv` could be confused with receiving a pubsub message. Maybe `change-msg-recv`?

src/gordon_gcp/plugins/janitor/gpubsub_publisher.py
```diff
@@ -238,16 +238,21 @@ def publish(self, message):
...
             try:
                 await self.publish(change_message)
+                await self.metrics.incr(
+                    'janitor-msg-publish', context=context)
```
econchick: Same here with the confusion over type of message.

smstone:
@spotify/alf I've addressed all comments. Can you PTAL?

src/gordon_gcp/plugins/janitor/authority.py
```diff
@@ -139,9 +143,13 @@ def __init__(self, config, crm_client, gce_client, rrset_channel=None,
     async def _get_projects(self):
         projects = await self._get_active_project_ids()
         project_blacklist = set(self.config.get('project_blacklist', []))
-        # TODO: emit a metric for all projects and a metric for
-        # project - blacklist.
-        return sorted(projects - project_blacklist)
+        sorted_projects = sorted(projects - project_blacklist)
+        context = {
+                'plugin': 'gceauthority'
+            }
```
econchick: Looks like some weird indentation?

PR #77, opened 2018-09-05
Fix TZ-aware/naive mismatch in consumer

ocsw:
https://github.com/spotify/gordon-gcp/blob/develop/src/gordon_gcp/plugins/service/event_consumer.py#L318 breaks because one object is TZ-aware and one is naive.  I thought the pubsub one was naive when I wrote it, but apparently not.

Fixes #

**Release note**:

```release-note
Fixed timezone bug in consumer.
```

### DIFF
--- a/tests/unit/plugins/service/test_event_consumer.py
+++ b/tests/unit/plugins/service/test_event_consumer.py
@@ -51,13 +51,19 @@ def raw_msg_data(creation_audit_log_data):
 
 
 @pytest.fixture
-def pubsub_msg(mocker, raw_msg_data):
+def publish_time():
+    # don't use in pubsub_msg because other things are assuming it's current
+    return datetime.datetime(
+        2018, 1, 1, 11, 30, 0, tzinfo=datetime.timezone.utc)
+
+
+@pytest.fixture
+def pubsub_msg(mocker, raw_msg_data, publish_time):
     mocker.patch(DATETIME_PATCH, conftest.MockDatetime)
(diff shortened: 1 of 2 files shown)


econchick:
@ocsw is it feasible to update the corresponding test(s) to catch this?

ocsw:
Yeah I realized after uploading.  _Should_ be simple, but it seems less so than I thought.

ocsw:
Oh wait, you meant to catch the problem directly?  Not without delving directly into the guts of the pubsub library.

tests/unit/plugins/service/test_event_consumer.py
```diff
@@ -363,12 +369,17 @@ def mock_create_gevent_msg(consumer, mocker, monkeypatch):
...
+                                     publish_time, raw_msg_data,
+                                     creation_audit_log_data,
                                      caplog, pubsub_msg, mock_get_and_validate,
                                      mock_create_gevent_msg):
     """Too-old message."""
-    mocker.patch(DATETIME_PATCH, conftest.MockDatetime)
-    pubsub_msg.publish_time -= datetime.timedelta(
+
+    def now(tz):
+        return publish_time
```
econchick: FWIW I think this could be done in a `lambda` directly in the `mocker.patch`

ocsw: Oh, good call.  Kept evolving.  I knew there was a way to make it even shorter. :)

PR #78, opened 2018-09-05
Dedupe instances by IP

ocsw:
Google can return both an instance that is in the middle of being deleted and one that is being created, with the same IP, so we need to dedupe them.

Fixes #

**Release note**:

```release-note
Fixed Google duplicate-IP problem.
```


ocsw:
I think that list of commits is because I reused a name?  Going to make a new branch.

PR #79, opened 2018-09-05
Dedupe instances by IP

ocsw:
Google can return both an instance that is in the middle of being deleted and one that is being created, with the same IP, so we need to dedupe them.

Fixes #

**Release note**:

```release-note
Fixed Google duplicate-IP problem.
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,4 +10,5 @@ git+git://github.com/GoogleCloudPlatform/google-cloud-python@861d51a20bddafdce84
 gordon-dns==0.0.1.dev6
 gordon-janitor==0.0.1.dev5
 jsonschema==2.6.0
+python-dateutil==2.7.3
 zope.interface==4.4.3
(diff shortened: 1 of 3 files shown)


ocsw:
@spotify/alf PTAL, duplicate IP problem

ocsw:
@gordon/alf applied Will's suggestions, PTAL

descent3me:
Is it insufficient to filter on the status of the instance when janitor talks to the GCE API?

descent3me:
Wouldn't the status be set to TERMINATED?

https://cloud.google.com/compute/docs/reference/rest/v1/instances/aggregatedList

ocsw:
It's also set to terminated when you stop an instance but don't delete it.  https://cloud.google.com/compute/docs/instances/stopping-or-deleting-an-instance

descent3me:
We probably don't want records for stopped instances either, right?

ocsw:
I think we do.  What if you take a box down for a few minutes and then bring it back up?  I don't think you want to wait 15+ minutes for it to reappear.

ocsw:
Also, I don't actually know currently what the statuses are and if they're actually logical.  We already know Google isn't really Doing The Right Thing in one way, I'm not sure I want to trust them in two.

ocsw:
How's this for an idea?  I just added logging.  We can run it for a while and see what the conflicts actually look like.  If they consistently have one instance in TERMINATED, we can junk this code and filter by status in the original query.

PR #80, opened 2018-09-06
Speed up reconciler

ocsw:
From watching janitor logs in realtime, I have figured out that we are spending a LOT of time in the list comprehensions in the reconciler that do the comparisons between desired and actual.  If we change these lookups to use dicts, I suspect we can shave a lot of time off.  (The existing comprehensions are O(n^2), the new version is O(n) because dicts are hashed.)

Fixes #

**Release note**:

```release-note
Sped up the reconciler.
```


ocsw:
@spotify/alf PTAL reconciler speedup

ocsw:
Cleared up comment based on verbal feedback from @descent3me

PR #86, opened 2018-09-11
Log aiohttp request call failures

ocsw:
We are seeing aiohttp request calls dying with timeouts.  This logs more info so we can figure out which ones are timing out.

Partial traceback:
```
  File "/usr/src/app/.venv/lib/python3.6/site-packages/gordon_gcp/plugins/janitor/authority.py", line 155, in _filter_results
    raise result
  File "/usr/src/app/.venv/lib/python3.6/site-packages/gordon_gcp/clients/gce.py", line 116, in list_instances
    responses = await self.list_all(url, params)
  File "/usr/src/app/.venv/lib/python3.6/site-packages/gordon_gcp/clients/_utils.py", line 51, in list_all
    response = await self.get_json(url, params=params)
  File "/usr/src/app/.venv/lib/python3.6/site-packages/gordon_gcp/clients/http.py", line 196, in get_json
    response = await self.request(method='get', url=url, **kwargs)
  File "/usr/src/app/.venv/lib/python3.6/site-packages/gordon_gcp/clients/http.py", line 146, in request
    method, url, **req_kwargs) as resp:
  File "/usr/src/app/.venv/lib/python3.6/site-packages/aiohttp/client.py", line 843, in __aenter__
    self._resp = await self._coro
  File "/usr/src/app/.venv/lib/python3.6/site-packages/aiohttp/client.py", line 465, in _request
    break
  File "/usr/src/app/.venv/lib/python3.6/site-packages/aiohttp/helpers.py", line 677, in __exit__
    raise asyncio.TimeoutError from None
concurrent.futures._base.TimeoutError
```

Fixes #

**Release note**:

```release-note
Add logging on aiohttp request call.
```

### DIFF
--- a/src/gordon_gcp/clients/http.py
+++ b/src/gordon_gcp/clients/http.py
@@ -142,42 +142,51 @@ class AIOConnection:
             method=method.upper(),
             url=url,
             kwargs=req_kwargs))
-        async with self._session.request(
-                method, url, **req_kwargs) as resp:
-            log_kw = {
-                'request_id': request_id,
-                'method': method.upper(),
-                'url': resp.url,
-                'status': resp.status,
-                'reason': resp.reason
-            }
-            logging.debug(_utils.RESP_LOG_FMT.format(**log_kw))
+        try:
+            async with self._session.request(method, url, **req_kwargs) as resp:
+                log_kw = {
+                    'request_id': request_id,
+                    'method': method.upper(),
+                    'url': resp.url,
+                    'status': resp.status,
+                    'reason': resp.reason
+                }
+                logging.debug(_utils.RESP_LOG_FMT.format(**log_kw))
 
-            if resp.status in REFRESH_STATUS_CODES:
-                logging.warning(f'[{request_id}] HTTP Status Code {resp.status}'
-                                f' returned requesting {resp.url}: '
-                                f'{resp.reason}')
-                if token_refresh_attempts:
-                    logging.info(f'[{request_id}] Attempting request to '
-                                 f'{resp.url} again.')
-                    return await self.request(
-                        method, url,
-                        token_refresh_attempts=token_refresh_attempts,
-                        request_id=request_id,
-                        **req_kwargs)
+                if resp.status in REFRESH_STATUS_CODES:
+                    logging.warning(
+                        f'[{request_id}] HTTP Status Code {resp.status}'
+                        f' returned requesting {resp.url}: {resp.reason}')
+                    if token_refresh_attempts:
+                        logging.info(
+                            f'[{request_id}] Attempting request to {resp.url} '
+                            'again.')
+                        return await self.request(
+                            method, url,
+                            token_refresh_attempts=token_refresh_attempts,
+                            request_id=request_id,
+                            **req_kwargs)
 
-                logging.warning(f'[{request_id}] Max attempts refreshing auth '
-                                f'token exhausted while requesting {resp.url}')
+                    logging.warning(
+                        f'[{request_id}] Max attempts refreshing auth token '
+                        f'exhausted while requesting {resp.url}')
 
-            # avoid leaky abstractions and wrap http errors with our own
-            try:
                 resp.raise_for_status()
-            except aiohttp.ClientResponseError as e:
-                msg = f'[{request_id}] Issue connecting to {resp.url}: {e}'
-                logging.error(msg, exc_info=e)
-                raise exceptions.GCPHTTPResponseError(msg, resp.status)
 
-            return await resp.text()
+                return await resp.text()
+        except aiohttp.ClientResponseError as e:
+            # bad HTTP status; avoid leaky abstractions and wrap HTTP errors
+            # with our own
+            msg = f'[{request_id}] HTTP error response from {resp.url}: {e}'
+            logging.error(msg, exc_info=e)
+            raise exceptions.GCPHTTPResponseError(msg, resp.status)
+        except exceptions.GCPHTTPResponseError as e:
+            # from recursive call
+            raise e
+        except Exception as e:
+            msg = f'[{request_id}] Request call failed: {e}'
+            logging.error(msg, exc_info=e)
+            raise exceptions.GCPHTTPError(msg)
 
     async def get_json(self, url, json_callback=None, **kwargs):
         """Get a URL and return its JSON response.
--- a/tests/unit/clients/test_http.py
+++ b/tests/unit/clients/test_http.py
@@ -179,7 +179,7 @@ async def test_request_max_request_attempts_reached(
         with pytest.raises(exceptions.GCPHTTPResponseError) as e:
             await client.request('get', conftest.API_URL)
 
-        e.match('Issue connecting to https://example.com/v1/foo_endpoint:')
+        e.match('HTTP error response from https://example.com/v1/foo_endpoint:')
 
     assert 2 == mock_refresh_token_called
     assert 9 == len(caplog.records)
@@ -243,6 +243,37 @@ async def test_request_with_valid_token(
     assert 0 == mock_refresh_token_called
     assert resp == resp_text
     assert 2 == len(caplog.records)
+
+
+@pytest.mark.asyncio
+async def test_request_with_other_exception(client, mocker, monkeypatch,
+                                            caplog):
+    """Correct exception is raised when the aiohttp request itself fails."""
+
+    async def mock_coro():
+        pass
+
+    monkeypatch.setattr(client._auth_client, 'refresh_token', mock_coro)
+    monkeypatch.setattr(client, 'valid_token_set', mock_coro)
+
+    monkeypatch.setattr('uuid.uuid4', lambda: 'FOO')
+
+    class AsyncContextManager:
+        def __init__(self, *args, **kwargs):
+            pass
+
+        async def __aenter__(self):
+            raise ValueError('BAR')  # randomly chosen
+
+        async def __aexit__(self, exc_type, exc, tb):
+            pass
+
+    monkeypatch.setattr(aiohttp.ClientSession, 'request', AsyncContextManager)
+
+    exp_msg = r'\[FOO\] Request call failed: BAR'
+    with pytest.raises(exceptions.GCPHTTPError, match=exp_msg):
+        await client.request(
+            'get', conftest.API_URL, token_refresh_attempts=0)
 
 
 def simple_json_callback(resp):


ocsw:
@spotify/alf PTAL more logging

descent3me:
+1 from me, but another pair of eyes wouldn't hurt since we were pairing and we aren't 100% sure what `release` and `close` do

src/gordon_gcp/clients/http.py
```diff
@@ -142,42 +142,50 @@ def __init__(self, auth_client=None, session=None):
...
-                logging.error(msg, exc_info=e)
-                raise exceptions.GCPHTTPResponseError(msg, resp.status)
-
-            return await resp.text()
+        try:
+            resp = await self._session.request(method, url, **req_kwargs)
+        except Exception as e:
+            msg = f'[{request_id}] Request call failed {resp.url}: {e}'
+            logging.error(msg, exc_info=e)
+            resp.close()
```
smstone: do we really need to attempt to close this given the request raised an exception?

ocsw: Well, it might remove the secondary exception in the logs.  We don't super much care I guess, just seemed like a thing to do.  Although I suppose it could set off its own crap.

smstone:
@descent3me @ocsw is this something we can test locally or in a test environment since you've said you're unsure of what these new methods do?

src/gordon_gcp/clients/http.py
econchick: Haven’t looked too closely at this patch yet, but doesn’t this need to close the session (rather than the response) since we’re no longer using the context manager? I would think it’s also a coroutine that needs to be awaited.

econchick: Haven’t looked too closely at this patch yet, but doesn’t this need to close the session (rather than the response) since we’re no longer using the context manager? I would think it’s also a coroutine that needs to be awaited.

ocsw: We were kinda unsure exactly what the deal is with all the release() and close() methods all over the place.  But we can also just drop this and let it error on process cleanup.

ocsw:
@smstone We're trying to figure out how to do that...

src/gordon_gcp/clients/http.py
descent3me: I don't think we really do, but why not be thorough?  There's other errors log a few lines down about an unclosed session when the program is shutting down, so at the very least it will make the log slightly cleaner.

```
gordon-janitor (1) ERROR: Unclosed connector
connections: ['[(<aiohttp.client_proto.ResponseHandler object at 0x7fcac0144940>, 1264.993380062)]']
connector: <aiohttp.connector.TCPConnector object at 0x7fcac21a23c8>
...
gordon-janitor (1) ERROR: Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x7fcac2135cf8>
```

descent3me:
@smstone Agreed that we should test, given that the docs are a little vague.

ocsw:
It also runs ok when I force an exception, but we still get cleanup errors on process exit.  I'm ok with removing the close in the except block.  @spotify/alf thoughts?

ocsw:
Turns out those errors are from the reconciler and the publisher, not the authority.

ocsw:
Original:
```
2018-09-09T17:05:31.651Z gordon-janitor (1) ERROR: Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x7f43270f6e80>
Exception ignored in: <coroutine object GDNSReconciler.run at 0x7f43270591a8>
...
```

Current:
```
Exception ignored in: <coroutine object GDNSReconciler.run at 0x7f69055b5830>
Traceback (most recent call last):
  File "/usr/src/app/.venv/lib/python3.6/site-packages/gordon_gcp/plugins/janitor/reconciler.py", line 356, in run
...
```

src/gordon_gcp/clients/http.py
```diff
@@ -142,42 +142,50 @@ def __init__(self, auth_client=None, session=None):
...
-                raise exceptions.GCPHTTPResponseError(msg, resp.status)
-
-            return await resp.text()
+        try:
+            resp = await self._session.request(method, url, **req_kwargs)
+        except Exception as e:
+            msg = f'[{request_id}] Request call failed {resp.url}: {e}'
+            logging.error(msg, exc_info=e)
+            resp.release()
+            raise e
```
econchick: Since we wrap aiohttp exceptions in our own (i.e. below in line 184 of the patch), I would want us to raise our own exception (i.e. `exceptions.GCPHTTPResponseError`) . Otherwise, the code calling this code will have to be updated to catch this mystery exception.

ocsw:
The cases we're trying to catch never get to that raise_for_status() bit.  They were raising exceptions within the setup of the 'async with request' line itself (__aenter__) and killing the program.  We can wrap them if we want, but we don't actually even know what all the possibilities are.  We were just trying to catch the exception long enough to log it, then let it kill the program like it was doing before.

econchick:
Sorry for being obtuse, but I'm still confused, [where in the `__aenter__`](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client.py#L854-L856) is it raising? It seems like we may be re-implementing the same functionality, unless I'm missing something?

w/r/t errors to raise, the wrapping is meant to take care of the unknown possibilities:

```py
try:
    resp = await request(...)
except Exception as e:
    msg = f'error doing something...  {e}'
    logging.error(msg)
    raise exceptions.GCPHTTPException(msg)
```

This way, whomever is calling `request` can just catch the `exceptions.GCPHTTPException` rather than not know how to handle an unwrapped aiohttp exception. Does that make sense?

But if we indeed do want it to actually kill the program on when this happens (which I question if we need to?), perhaps we should insist on a hard exit, i.e. `sys.exit(1)`?

src/gordon_gcp/clients/http.py
econchick: Also I just realized, the `resp` variable will not be available to you for the `msg` string nor `release` it since that's what raised the exception. That makes me wonder what you exactly you see when you're testing this locally, i.e. are you ever able to get to this point.

ocsw: Thaaaaat... is a good point.  I tested it by raising an exception after that line.  The good news is we don't need it for the msg as long as we have the request ID.

ocsw:
When I said __aenter__ was raising I was mostly just pointing out that the `with` line itself was raising.  But for clarity, we call request() on our ClientSession object.  That creates a wrapped coro of _request(), and the __aenter__ of that wrapper object is what the `with` calls.  It awaits the coro and _request raises on a timeout.

In terms of our wrappers, I guess I'm ok with your plan.  It will still kill the program in this case, from _filter_results() in the authority.  That's actually what was happening in the first place.  I kinda misspoke.  The timeout exception gets returned from asyncio.gather() and the filter blows up, it doesn't die directly when it hits the exception because of the gather.

econchick:
I’m not sure I understand the answer to what this is trying to solve. I can follow this and aiohttp’s code. It seems to me that the code changes here just reimplement what aiohttp’s private api is doing; that is, without context managers. For instance, this probably will still have an open or otherwise not cleaned up connection at the time an exception happens since aiohttp’s response object and request context object handles that for us in a context manager. For that case you’d need to have that clean up part in a “finally” clause after the try/except. So then, why not wrap the whole “async with” statement in a try/except? So we can let aiohttp take care of all the mess it created.

ocsw:
Ah, I see what you mean.  We looked at wrapping the whole thing, but we were concerned because then our exception block would also catch anything raised inside the context manager block.  I suppose we could just reraise GCPHTTPResponseError with no processing, and only log the rest.  Took me a bit to realize what you were pointing at.  I'll make another version.

ocsw:
@econchick how's this?

src/gordon_gcp/clients/http.py
```diff
@@ -142,42 +142,52 @@ def __init__(self, auth_client=None, session=None):
...
+                # avoid leaky abstractions and wrap http errors with our own
+                try:
+                    resp.raise_for_status()
+                except aiohttp.ClientResponseError as e:
+                    msg = f'[{request_id}] Issue connecting to {resp.url}: {e}'
+                    logging.error(msg, exc_info=e)
+                    raise exceptions.GCPHTTPResponseError(msg, resp.status)
+
+                return await resp.text()
+        except Exception as e:
```
econchick: What do you think of this:

```py
try:
    async with self._session.request(method, url, **req_kwargs) as resp:
        # logic
       resp.raise_for_status()  # no wrapping here
       return await resp.text()
except aiohttp.ClientResponseError as e:
    # move the raise_for_status error handling to here; 
    # should have access to the `resp` variable
except Exception as e:
    # all other exception handling (no access to `resp` variable)
```

In essence, moving the inner `try`/`except` to the outer `try`/`except`.

ocsw: I considered that, but is it separating functions too much?  The handler has stuff about status that might be easier to follow where it is now.

ocsw: I'm open to either way though.

descent3me: I see the code's changed a bit since the last time I saw it, but if @econchick 's comment is agains the code as it is now, I prefer her suggestion, since it reduces the nesting of the try blocks and having to check `isinstance`

ocsw: ok, done.

ocsw:
Fixed the test failure.  I hadn't been planning to write a new test for this exception, but I suppose I should.

ocsw:
@econchick OK, I think everything is taken care of?

src/gordon_gcp/clients/http.py
```diff
@@ -142,42 +142,51 @@ def __init__(self, auth_client=None, session=None):
...
-            return await resp.text()
+                return await resp.text()
+        except aiohttp.ClientResponseError as e:
+            # bad HTTP status; avoid leaky abstractions and wrap HTTP errors
+            # with our own
+            msg = f'[{request_id}] HTTP error response from {resp.url}: {e}'
+            logging.error(msg, exc_info=e)
+            raise exceptions.GCPHTTPResponseError(msg, resp.status)
+        except exceptions.GCPHTTPResponseError as e:
+            # from recursive call
```
econchick: Confused, from where would `exceptions.GCPHTTPResponseError` be raised?

tests/unit/clients/test_http.py
```diff
@@ -245,6 +245,36 @@ def client(mocker, auth_client):
...
+        client, mocker, monkeypatch, caplog):
+    """Correct exception is raised when the aiohttp request itself fails."""
+
+    async def mock_refresh_token():
+        pass
+
+    monkeypatch.setattr(
+        client._auth_client, 'refresh_token', mock_refresh_token)
+
+    async def mock_valid_token_set():
```
econchick: These two mocks could be generalized, since they do the same thing, i.e. `async def mock_coro`.

tests/unit/clients/test_http.py
```diff
@@ -245,6 +245,36 @@ def client(mocker, auth_client):
...
+        pass
+
+    monkeypatch.setattr(client, 'valid_token_set', mock_valid_token_set)
+
+    mock_aiohttp_request = mocker.Mock()
+    mock_aiohttp_request.side_effect = ValueError  # randomly chosen
```
econchick: Doesn't it need to be instantiated? i.e `ValueError('not a valid value')`?

tests/unit/clients/test_http.py
```diff
@@ -245,6 +245,36 @@ def client(mocker, auth_client):
...
+    monkeypatch.setattr(
+        client._session, 'request', mock_aiohttp_request
```
econchick: I thought this would be a coroutine?

src/gordon_gcp/clients/http.py
ocsw: The first time we call the function and get a 401, it calls itself.  That call still returns a 401, so it raises that exception.  The first call then gets it, and it hits these exception blocks in the first call.

tests/unit/clients/test_http.py
ocsw: Good point.

tests/unit/clients/test_http.py
ocsw: Exceptions default to ('') iirc.  In any case, it works.  But I can add that to further test the log message.

tests/unit/clients/test_http.py
ocsw: That's a good point.  I guess it works because you can await a non-awaitable and it just does nothing.  Maybe I should make an async function just for clarity?

tests/unit/clients/test_http.py
ocsw: Oh wait, you mean because the 'side effect' is a class.  The mock library checks if it's an exception and instantiates it for you.

tests/unit/clients/test_http.py
ocsw: @econchick It turns out that doing it this way is actually much simpler.  If you make it a coro, you have to implement __aenter__ and __aexit__, while this way seems not to actually apply the context manager protocol.  Thoughts?

ocsw: It also turns out that the mock library won't let you mock a magic function like __aenter__.

econchick: ftr mock doesn't support async stuff at all (maybe in 3.7? an unconfirmed rumor).

ocsw: OK, I've added a more thorough solution that actually also matches what aiohttp does better.

ocsw: @econchick re mock, I also discovered that if you try to add a method that starts with __, it won't let you.  Or maybe surrounded by __, didn't check the exact criteria.

PR #87, opened 2018-09-12
Add additional octet test cases for get_managed_zone

RochesterinNYC:
- Add examples to method docstring
- This makes it more clear that the get_managed_zone can handle reverse zones with numbers of octets other than just 2


ocsw:
Might want to explicitly say in the docstring that you do need at least 2?  Otherwise +1

descent3me:
+1

PR #93, opened 2018-10-08
[WIP] Allow custom managed zone prefix

descent3me:
Allows configuring a custom prefix for managed zone names

Fixes #

**Release note**:

```release-note
Allow supplying custom managed zone name prefix.
```


ocsw:
Maybe this is coming later, but more explanation of why you would want to do that would be helpful.  And maybe we should discuss other solutions for the testing use case?

descent3me:
Hi, @spotify/alf This is ready for review.

@ocsw As in the test example, the purpose for prepending all zones with an arbitrary string could be to have multiple sets of zones in the same project, some used for production and some for various testing purposes, or to simply name all zones in a way that makes sense to the user of Gordon.  For the former @econchick and I discussed the use of labels. I think this is a fine idea, but it requires a larger change since none of the API endpoints we're using support accessing a zone in this way. It would be possible to have the DNS client query for all available zones during startup, and manually filter on label (there is nothing in HTTP API for filtering while querying) to store references to them locally and then some changes to `get_managed_zone` to do a lookup among those stored zones rather than a string operation, but this is a larger functional change without a significant benefit to the simple fix that I can see.

src/gordon_gcp/clients/gdns.py
```diff
@@ -102,10 +102,13 @@ class GDNSClient(http.AIOConnection):
     REVERSE_PREFIX = 'reverse-'
 
     def __init__(self, project=None, auth_client=None, api_version='v1',
-                 session=None):
+                 session=None, default_zone_prefix=None):
         super().__init__(auth_client=auth_client, session=session)
         self.project = project
         self._base_url = f'{self.BASE_URL}/{api_version}/projects/{project}'
+        prefix = f'{default_zone_prefix}-' if default_zone_prefix else ''
```
ocsw: I would leave the - out and have the user put it in the config.  If you really want to leave it, please document that it's added.

ocsw:
Re explanation, I meant in the PR description.

One issue, otherwise +1.

src/gordon_gcp/clients/gdns.py
descent3me: I thought it was consistent enough with our naming of the zones that it would be unsurprising (an earlier version of this PR actually removed it from the REVERSE_PREFIX) and only added it here as well, since I don't really see that as part of the prefix itself), but I can update the doc.

ocsw:
had to pull out the doc fix https://github.com/spotify/gordon-gcp/pull/95

descent3me:
@spotify/alf Can I get a quick re-review? I added the requested docs line and rebased out `Fix Google APIs sphinx redirect link 7331a4b`

docs/config-janitor.rst
```diff
@@ -47,12 +47,17 @@ Any configuration key/value listed here may also be used in the specific plugin
 
     `Optional`: Timeout in seconds for how long each plugin should wait for outstanding tasks (e.g. processing remaining message from a channel) before cancelling. This is only used when a plugin has received all messages from a channel, but may have work outstanding. Defaults to ``60``.
 
+.. option:: default_zone_prefix="STR"
+
+    `Optional`: Prefix associated with Google managed zone names, prepended with a '-' to the generated name.
```
ocsw: I think the phrasing is a little unclear.  'prepended to the generated name with a - in between' and then an example, maybe?

descent3me: I think the between makes it more confusing, but I have added an example.

ocsw:
+1

PR #94, opened 2018-10-09
Filter out instances with state = TERMINATED

ocsw:
Since we're seeing periodic janitor failures with the TERMINATED filter in config, let's try filtering them out in code.

Fixes #

**Release note**:

```release-note
Filter out instances with state = TERMINATED in the authority.
```

### DIFF
--- a/src/gordon_gcp/plugins/janitor/authority.py
+++ b/src/gordon_gcp/plugins/janitor/authority.py
@@ -174,8 +174,15 @@ class GCEAuthority:
             coros.add(coro)
 
         all_results = await asyncio.gather(*coros, return_exceptions=True)
+        all_results = self._filter_results(all_results)
 
-        return self._filter_results(all_results)
+        instances = []
+        for instance in all_results:
+            if instance['status'] == 'TERMINATED':
+                logging.info(f'Skipping terminated instance: "{instance}"')
+                continue
+            instances.append(instance)
+        return instances
 
     def _create_instance_rrset(self, instance):
         ip = instance['networkInterfaces'][0]['accessConfigs'][0]['natIP']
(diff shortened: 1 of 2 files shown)


ocsw:
@spotify/alf PTAL

ocsw:
I got that doc problem locally too but I thought it might be a network issue.  Looks like something we're relying on vanished?  Although I was able to download _something_ when I went to the URL, then it gave me a 404 page?

ocsw:
(@descent3me fixed that in another PR)

src/gordon_gcp/plugins/janitor/authority.py
```diff
@@ -174,8 +174,15 @@ def _filter_results(self, results):
...
-
-        return self._filter_results(all_results)
+        all_results = self._filter_results(all_results)
+
+        instances = []
+        for instance in all_results:
+            if instance['status'] == 'TERMINATED':
+                logging.info(f'Skipping terminated instance: "{instance}"')
+                continue
+            instances.append(instance)
```
econchick: Any harm in moving this up to the `_filter_results` method? We seem to be doing some skipping up there.

ocsw: I was thinking this is better grouping of functionality - filter results filters the results themselves, get instances gets (filtered) instances?

tests/unit/plugins/janitor/test_authority.py
```diff
@@ -232,3 +236,11 @@ def test_create_msgs_bad_json(caplog, fake_authority):
     results = fake_authority._create_msgs([partial_instance])
     assert [] == results
     assert 1 == len(caplog.records)
+
+
+# test_get_instances_terminated(mocker, authority_config, metrics,
```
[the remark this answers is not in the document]
ocsw: oops

PR #96, opened 2018-10-15
Lower minimum TTL in event schema to match reality

ocsw:
We have DNS records with TTLs below 300 which were being dropped by this schema.

Fixes #

**Release note**:

```release-note
Lower minimum TTL for event schema to 60
```


ocsw:
@RochesterinNYC PTAL bugfix

PR #101, opened 2018-11-16
Add project whitelist for janitor authority

william2771:
Add a project whitelist to specify a project to run janitor against, makes testing easier.

Fixes #

If the project_whitelist is set, the authority will only look at those projects for fetching instances. Blacklist and active projects are ignored.

**Release note**:

```release-note
Add a project whitelist to Janitor Authority
```

### DIFF
--- a/docs/config-janitor.rst
+++ b/docs/config-janitor.rst
@@ -101,6 +101,10 @@ Additional plugin-specific configuration is needed:
 
     `Optional`: List of unique, user-assigned project IDs (``projectId``) that will be ignored when fetching projects.
 
+.. option:: project_whitelist=["STR","STR"]
+
+    `Optional`: List of unique, user-assigned project IDs (``projectId``) that will be used to fetch instances. If set, janitor will only look at these projects, it will not fetch active projects and the project_blacklist will be ignored.
+
 .. option:: instance_filter="STR"
 
     `Optional`: String used to filter instances by instance attributes. It is passed directly to GCE's `v1.instances.aggregatedList <https://cloud.google.com/compute/docs/reference/rest/v1/instances/aggregatedList>`_ endpoint's `filter` parameter.
--- a/docs/config-janitor.rst
+++ b/docs/config-janitor.rst
@@ -89,7 +89,7 @@ Additional plugin-specific configuration is needed:
 
     Note: this is separate from Google's 'managed zone' names.  Google uses custom string names with specific `requirements <https://cloud.google.com/dns/api/v1/managedZones#resource>`_ for storing records. Gordon requires that managed zone names be based on DNS names. For all domains, remove the trailing dot and replace all other dots with dashes.  For reverse records, then use only the two most significant octets, prepended with 'reverse-'.  (E.g. ``foo.bar.com.`` -> ``foo-bar-com`` and ``0.168.192.in-addr.arpa.`` -> ``reverse-168-192.``)
 
-.. option:: metadata_blackklist=[["STR","STR"],["STR","STR"]]
+.. option:: metadata_blacklist=[["STR","STR"],["STR","STR"]]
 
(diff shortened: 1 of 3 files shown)


src/gordon_gcp/plugins/janitor/authority.py
```diff
@@ -143,6 +143,9 @@ def __init__(self, config, metrics, crm_client, gce_client,
     async def _get_projects(self):
         projects = await self._get_active_project_ids()
         project_blacklist = set(self.config.get('project_blacklist', []))
+        project_whitelist = self.config.get('project_whitelist')
+        if project_whitelist:
+            projects &= set(project_whitelist)
```
descent3me: This is interesting, but I think the Janitor is still doing more work than it needs to. If someone adds projects to the whitelist, we could just skip the fetching of active project IDs altogether.

tests/unit/plugins/janitor/test_authority.py
```diff
@@ -162,6 +162,84 @@ def _create_fake(klass, *args, **kwargs):
         'rrsets', len(expected_rrsets), context=context)])
 
 
+@pytest.mark.asyncio
+async def test_whitelist_project_publishes_msg(mocker, authority_config,
+                                               get_gce_client, create_mock_coro,
+                                               instance_data, metrics):
```
descent3me: This also looks like a very thorough test (I haven't gotten all the way through it yet), but for such a small change I think it would be acceptable to do a unit test on the `_get_projects` method. For me it would be more clear to see what was changed and what to expect from the new behaviour.

src/gordon_gcp/plugins/janitor/authority.py
descent3me: I think project_whitelist is fine. If that happens, I think it's okay for the Janitor to fail, as long as we document how the parameter works in https://github.com/spotify/gordon-gcp/blob/develop/docs/config-janitor.rst

william2771: I'll update the functionality and the doc 👍

tests/unit/plugins/janitor/test_authority.py
```diff
@@ -232,3 +232,31 @@ def test_create_msgs_bad_json(caplog, fake_authority):
     results = fake_authority._create_msgs([partial_instance])
     assert [] == results
     assert 1 == len(caplog.records)
+
+
+@pytest.mark.parametrize('whitelist,expected', [
+    (['whitelisted_1', 'whitelisted_1'], ['whitelisted_1', 'whitelisted_1']),
```
descent3me: It might be a good idea to use two unique names to illustrate what's happening. I realize this will work (and since it's parametrized we could leave this one around too to show that uniqueness isn't important), but it's harder to immediately tell what's going on from looking at this case.

docs/config-janitor.rst
```diff
@@ -101,6 +101,10 @@ Additional plugin-specific configuration is needed:
 
     `Optional`: List of unique, user-assigned project IDs (``projectId``) that will be ignored when fetching projects.
 
+.. option:: project_whitelist=["STR","STR"]
+
+    `Optional`: List of unique, user-assigned project IDs (``projectId``) that will be used to fetch instances. If set, janitor will only look at these projects, it will not fetch active projects and the project_blacklist will be ignored.
```
descent3me: I'm not familiar enough with the rst format to know if there's a way to refer back to another parameter for `project_blacklist`, but if so that might be nice. Otherwise maybe quotes or backticks or some other emphasis should be around it to show it's also a config parameter?

tests/unit/plugins/janitor/test_authority.py
william2771: derp forgot to rename my test values...

PR #106, opened 2019-02-13
Add/adjust metrics

descent3me:
We've been running gordon and gordon janitor for some time now, and noticed some metrics were missing or not as flexible as we hoped. 

Fixes #

@spotify/alf 

**Release note**:

```release-note
Add and adjust some janitor and service metrics.
```

### DIFF
--- a/src/gordon_gcp/plugins/janitor/gpubsub_publisher.py
+++ b/src/gordon_gcp/plugins/janitor/gpubsub_publisher.py
@@ -247,14 +247,15 @@ class GPubsubPublisher:
                 break
             context = base_context.copy()
             context['action'] = change_message['action']
-            await self.metrics.incr('change-msg-recv', context=context)
             try:
                 await self.publish(change_message)
-                await self.metrics.incr(
-                    'change-msg-publish', context=context)
+                context['result'] = 'published'
             except Exception as e:  # todo
                 logging.error('Exception while trying to publish message to '
                               f' pubsub: {e}')
+                context['result'] = 'error'
+
+            await self.metrics.incr('change-message', context=context)
 
         await self.cleanup()
         await timer.stop()
--- a/src/gordon_gcp/plugins/janitor/gpubsub_publisher.py
+++ b/src/gordon_gcp/plugins/janitor/gpubsub_publisher.py
@@ -238,13 +238,15 @@ class GPubsubPublisher:
         Once ``None`` is received from the channel, finish processing
         records and clean up any outstanding tasks.
         """
-        context = {'plugin': 'gpubsub-publisher'}
-        timer = self.metrics.timer('plugin-runtime', context=context)
+        base_context = {'plugin': 'gpubsub-publisher'}
+        timer = self.metrics.timer('plugin-runtime', context=base_context)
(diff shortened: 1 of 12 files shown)


src/gordon_gcp/plugins/janitor/reconciler.py
```diff
@@ -69,6 +70,34 @@
...
+    """
+    # TODO (lynn): This will be moved to a common package to be shared
+    #   between all of gordon* packages. It will also make use of attrs
+    #   ability to optionally validate upon creation.
+    name = attr.ib(type=str)
+    type = attr.ib(type=str)
+    rrdatas = attr.ib(type=list)
+    kind = attr.ib(type=str, default='dns#resourceRecordSet')
+    ttl = attr.ib(type=int, default=300)
+    source = attr.ib(type=str, default=None)
```
descent3me: This line was added in this PR. The rest is just moved from the DNS client.

descent3me:
Here are examples of the new metrics using the log metric relay.

In gordon core:
```
2019-02-14T20:23:19.581Z gordon[1]: [message-validated] value: 6 context: {'plugin': 'event-consumer', 'schema': 'audit-log'}
```

In gordon janitor:
```
2019-02-14T20:37:07.386Z gordon-janitor (1) INFO: [rrsets-handled] value: 1 context: {'plugin': 'reconciler', 'action': 'additions', 'source': 'gceauthority'}
...
2019-02-14T20:37:07.391Z gordon-janitor (1) INFO: [change-message] value: 25 context: {'plugin': 'gpubsub-publisher', 'action': 'additions', 'result': 'published'}
2019-02-14T20:37:07.392Z gordon-janitor (1) INFO: [change-message] value: 26 context: {'plugin': 'gpubsub-publisher', 'action': 'deletions', 'result': 'published'}

```

src/gordon_gcp/plugins/janitor/gpubsub_publisher.py
```diff
@@ -247,14 +247,15 @@ def publish(self, message):
...
-            await self.metrics.incr('change-msg-recv', context=context)
             try:
                 await self.publish(change_message)
-                await self.metrics.incr(
-                    'change-msg-publish', context=context)
+                context['result'] = 'published'
             except Exception as e:  # todo
                 logging.error('Exception while trying to publish message to '
                               f' pubsub: {e}')
+                context['result'] = 'error'
```
RochesterinNYC: Does this metric/context include enough information about the error/exception that occurred ?

descent3me: I think so, because we can (should?) use the metric to know that *something* happened, but use the logs to debug, rather than trying to use a metric to work for both. Since we don't have any expected error causes, the only thing we could use would be something generic like class name, but I'm not sure how useful something like "NameError" or "KeyError" would be on it own. We'd probably still end up needing to check the logs to debug.

RochesterinNYC: :+1: That makes sense to me from an operations point of view.

PR #108, opened 2019-02-19
[reconciler] remove attrs and use sets for comparison

descent3me:
This PR contains a fix for the reconciler bug accidentally merged in with https://github.com/spotify/gordon-gcp/pull/106

There are three main components to the PR
1. Adjust `test_reconciler.validate_rrsets_by_zone` to actually fail based off the previous PR.
2. Drop use of `attrs` for `reconciler.ResourceRecordSet` in favor of a simple object where we control which attributes are used for comparison. It adds a few more lines of boilerplate code, but I feel it's easier to understand and control.
3. Simplify the reconciliation logic to remove extra list iterations and use sets instead.

Within those three changes there are some subtler changes. 
- Use tuple internally in `reconciler.ResourceRecordSet.rrdatas` and update the docstring to indicate rrdatas should be a iterables of strings (not possibly unhashable objects). This should be fine as tuple also json serializes as an array ([adjusted a test in publisher](https://github.com/spotify/gordon-gcp/pull/108/files#diff-2cdc114177c7a5745a4de1e484e0a9dbR85) to illustrate) and string is consistent with the Google Cloud DNS API (https://cloud.google.com/dns/docs/reference/v1/resourceRecordSets#resource) and needed to avoid unnecessary complexity for hashing.  
- No longer drop conflicting deletes. Previously if a record of same name and type were to be added and deleted from a zone, we would drop the deletion, because the gdns publisher in Gordon has some logic for deleting conflicting records. I think this was done to save sending an unnecessary pubsub message. However it should be relatively infrequent (say IP reuse) and adds complexity to the reconciliation process. Instead I have chosen to send the deletions first, so we can be sure that the deletion happens before the addition.

Fixes # https://github.com/spotify/gordon-gcp/pull/106

@spotify/alf 

**Release note**:

```release-note
Fix reconciler bug introduced in v0.0.1.dev30.
```

### DIFF
--- a/tests/unit/plugins/janitor/test_reconciler.py
+++ b/tests/unit/plugins/janitor/test_reconciler.py
@@ -18,7 +18,6 @@ import asyncio
 from unittest import mock
 
 import aiohttp
-import attr
 import pytest
 
 from gordon_gcp.clients import auth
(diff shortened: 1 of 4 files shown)


tests/unit/plugins/janitor/test_reconciler.py
```diff
@@ -259,29 +255,34 @@ def extra_rrset():
...
             for record in [rrsets[0], soa_ns_rrsets[2]]
-        ]
-        expected_extra_rrsets = [
+        )
+        expected_extra_rrsets = set(
             reconciler.ResourceRecordSet(**record)
-            for record in [extra_rrset]
-        ]
+            for record in [return_rrsets[0], extra_rrset]
+        )
```
descent3me: This is the test that was updated to fail.  It seems we were reusing the same objects in many parts of the test and none of the objects had "source".  I first updated the test to reuse less, added "source" to the input and expected rrsets only (not to those returned from the GCE API) and ensure it failed.

After updating the API, I also had to adjust the expected returned extra rrsets here because of the decision to not drop deletions for updates, and replace lists with sets where necessary.

descent3me:
@spotify/alf ping! I just rebased off the latest in develop ^

PR #109, opened 2019-02-22
Set google-cloud-pubsub to latest released version

bcleenders:
Bumping google-cloud-pubsub library to avoid depending on a non-release version.


descent3me:
The publish time issue was fixed in July https://github.com/googleapis/google-cloud-python/pull/5633

descent3me:
+1

PR #110, opened 2019-02-26
[event_consumer] kill Gordon on repeated subscribe failure

bcleenders:
If Gordon cannot create a pubsub subscription, there is no point in continuing running.

Add logic for retrying, and exiting if it fails repeatedly to create a subscription.


descent3me:
+1

PR #112, opened 2019-02-28
Feature/improve source metric

descent3me:
This changes the "source" of missing records from "unknown" to "gdns" in the "rrsets-handled" metric, which is easier to understand.  I also added a couple tests to get to 100% test coverage in reconciler, and adjusted an existing test to show how source is handled when there is a duplicate record from another authority.

Fixes #

@spotify/alf 

**Release note**:

```release-note
Change  "source" of missing records from "unknown" to "gdns" in "rrsets-handled" metric.
```


descent3me:
Thanks @william2771!

PR #113, opened 2020-04-14
Fix audit log regex

ocsw:
Gordon is throwing errors like this:

2020-04-14T20:53:59.989+00:00 10.99.0.2 gordon[1]: Given message was not valid against the schema "audit-log": 'projects/foo/zones/zzz/instances/bar' does not match '^projects/[0-9]/zones/[a-z\\-0-9]/instances/[a-z\\-0-9]+'

because the audit log has some projects with names instead of numeric IDs.

### DIFF
--- a/src/gordon_gcp/schema/schemas/audit-log.schema.json
+++ b/src/gordon_gcp/schema/schemas/audit-log.schema.json
@@ -23,7 +23,7 @@
                     "examples": [
                         "projects/a-project-id/zones/a-zone-name/instances/an-instance-name"
                     ],
-                    "pattern": "^projects/[0-9]+/zones/[a-z\\-0-9]+/instances/[a-z\\-0-9]+"
+                    "pattern": "^projects/([0-9]+|[a-z][a-z\\-0-9]{5,29})/zones/[a-z\\-0-9]+/instances/[a-z\\-0-9]+"
                 }
             },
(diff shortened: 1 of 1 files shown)


src/gordon_gcp/schema/schemas/audit-log.schema.json
```diff
@@ -23,7 +23,7 @@
                     "examples": [
                         "projects/a-project-id/zones/a-zone-name/instances/an-instance-name"
                     ],
-                    "pattern": "^projects/[0-9]+/zones/[a-z\\-0-9]+/instances/[a-z\\-0-9]+"
+                    "pattern": "^projects/([0-9]+|[a-z][a-z\\-0-9]{5,29})/zones/[a-z\\-0-9]+/instances/[a-z\\-0-9]+"
```
descent3me, line 26: I'd be willing to be more lenient for easier readability (and potentially for if GCP changes their requirements), like `[a-z\\-0-9]+`. I'm not sure we gain much by being so specific.

ocsw: I mean, legit.  But those aren't the same because of the initial letter.  Shrug.

descent3me:
Update: this fixed pytest's problem (second build step), but there seem to still be docs' problems (third build step)

PR #114, opened 2020-04-16
Fix build issues

descent3me:
Fix the build error in https://github.com/spotify/gordon-gcp/pull/113

Fixes #
 Build issues in https://github.com/spotify/gordon-gcp/pull/113
Also reorders py.test and flake8 because I think it's handy for local debugging.
**Release note**:

```release-note

```


descent3me:
Update: this fixed pytest's problem (second build step), but there seem to still be docs' problems (third build step)

ocsw:
Personally, I'd prefer having flake8 first, but I guess we've made a similar change elsewhere recently...

PR #116, opened 2020-04-20
Fallback to default token URI if not set

reneweb:

When using default compute engine credentials (i.e. default application credentials on GCE / GKE) the object won't contain a  `_token_uri` attribute. This PR provides a fallback to use the default token uri as defined here: https://github.com/googleapis/google-auth-library-python/blob/master/google/auth/compute_engine/credentials.py#L120

**Release note**:

```release-note

```


ocsw:
+1

ocsw:
Will need to release a package

reneweb:
@ocsw Not too familiar how this works, but seems like we first would merge this one and then have a separate PR for the release?

PR #118, opened 2020-04-21
Refresh function

reneweb:
Follow up from: https://github.com/spotify/gordon-gcp/pull/116 
This change adds support for compute engine credentials while still keeping the http calls async. To get a refresh token from compute engine credentials a GET request to the metadata server needs to be made rather than a POST to the token one.

### DIFF
--- a/src/gordon_gcp/clients/auth.py
+++ b/src/gordon_gcp/clients/auth.py
@@ -52,11 +52,14 @@ To use:
 import asyncio
 import json
 import logging
+import os
 import urllib.parse
 import uuid
 
 import aiohttp
 from google import auth as gauth
+from google.auth import compute_engine
+from google.auth import environment_vars
 from google.oauth2 import _client
 from google.oauth2 import service_account
 
@@ -101,6 +104,13 @@ class GAuthClient:
         self.token = None
         self.expiry = None  # UTC time
 
+        if isinstance(self.creds, compute_engine.credentials.Credentials):
+            self._refresh_token_f = \
+                self._refresh_token_using_compute_credentials
+        else:
+            self._refresh_token_f = \
+                self._refresh_token_using_service_account_credentials
+
     def _load_keyfile(self, keyfile):
         if not keyfile:
             return None
@@ -141,10 +151,7 @@ class GAuthClient:
         return session
 
     def _setup_token_request(self):
-        if hasattr(self.creds, '_token_uri'):
-            url = self.creds._token_uri
-        else:
-            url = _utils.DEFAULT_TOKEN_URI
+        url = self.creds._token_uri
 
         headers = _utils.DEFAULT_REQUEST_HEADERS.copy()
         headers.update(
@@ -178,14 +185,31 @@ class GAuthClient:
                 specifically a :exc:`.GCPHTTPResponseError`, if the
                 exception is associated with a response status code.
         """
+        await self._refresh_token_f()
+
+    async def _refresh_token_using_compute_credentials(self):
+        metadata_host = os.getenv(environment_vars.GCE_METADATA_ROOT,
+                                  'metadata.google.internal')
+        url = (f'http://{metadata_host}/computeMetadata/'
+               'v1/instance/service-accounts/'
+               f'{self.creds._service_account_email}/token')
+        headers = {'metadata-flavor', 'google'}
+        await self._execute_refresh_token_request(url, 'GET', headers)
+
+    async def _refresh_token_using_service_account_credentials(self):
         url, headers, body = self._setup_token_request()
+        await self._execute_refresh_token_request(url, 'POST', headers, body)
+
+    async def _execute_refresh_token_request(
+            self, url, method, headers, body=None):
         request_id = uuid.uuid4()
         logging.debug(_utils.REQ_LOG_FMT.format(
-            request_id=request_id, method='POST', url=url, kwargs=None))
-        async with self._session.post(url, headers=headers, data=body) as resp:
+            request_id=request_id, method=method, url=url, kwargs=None))
+        async with self._session.request(
+                method, url, headers=headers, data=body) as resp:
             log_kw = {
                 'request_id': request_id,
-                'method': 'POST',
+                'method': method,
                 'url': resp.url,
                 'status': resp.status,
                 'reason': resp.reason,
--- a/tests/unit/clients/test_auth.py
+++ b/tests/unit/clients/test_auth.py
@@ -23,6 +23,7 @@ import aiohttp
 import pytest
 from aioresponses import aioresponses
 from google import auth as gauth
+from google.auth import compute_engine
 from google.oauth2 import _client as oauth_client
 from google.oauth2 import credentials
 from google.oauth2 import service_account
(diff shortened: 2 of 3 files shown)


reneweb:
Still going to test a bit before merging

reneweb:
Fixed & added tests

reneweb:
I've changed it to manually refresh the token when using compute engine credentials

src/gordon_gcp/clients/auth.py
```diff
@@ -178,14 +178,38 @@ async def refresh_token(self):
...
+        else:
+            await self._refresh_token_using_service_account_credentials()
+
+    async def _refresh_token_using_compute_credentials(self):
+        metadata_host = os.getenv(environment_vars.GCE_METADATA_ROOT,
+                                  'metadata.google.internal')
+        url = "http://{0}/" \
+              "computeMetadata/v1/instance/service-accounts/{1}/token".format(
+                metadata_host,
+                self.creds._service_account_email)
```
descent3me, line 193: - on this line and others `"` should be `'` (by convention, code should be consistent in its choice of single vs double quote, so we went with single, besides the case of nested quotes, for example `x = log.error('an error with "something"')` )
- Python 3 also has format strings, which can make code a little more readable. We also use parentheses for multiline strings. For example this could become (hopefully I got the whitespace right 🤞)
```py3
        url =  (f'http://{metadata_host}/computeMetadata/'
                'v1/instance/service-accounts/'
                f'{self.creds._service_account_email}/token')
```

Here's the style guide we're following for reference: https://google.github.io/styleguide/pyguide.html

src/gordon_gcp/clients/auth.py
```diff
@@ -178,14 +178,38 @@ async def refresh_token(self):
                 specifically a :exc:`.GCPHTTPResponseError`, if the
                 exception is associated with a response status code.
         """
+        if isinstance(
+                self.creds, compute_engine.credentials.Credentials):
```
descent3me, line 182: In general, it is good practice to try to minimize use of `isinstance` (some adage similar to "check actions, not types"). It also seems like this probably isn't something that will change over the lifetime of the client, could the appropriate function could be set while setting up the credentials from the init method rather than on every refresh request?

src/gordon_gcp/clients/auth.py
reneweb: - yep, I missed to make those single quotes
- Will change to format strings

src/gordon_gcp/clients/auth.py
reneweb: Yep, I am aware that `isinstance` is not best practice in python. I could check on the existence of all the fields like `client_id`, `client_secret` and stuff, but even then the real differentiating factor is where the credentials are from (service account file, compute engine, etc) and that information is just in the type as far as I can see.
I can move the check in the init function and then just assign the function to some field and call it here.

src/gordon_gcp/clients/auth.py
```diff
@@ -101,6 +104,13 @@ def __init__(self, keyfile=None, scopes=None, session=None, loop=None):
         self.token = None
         self.expiry = None  # UTC time
 
+        if isinstance(self.creds, compute_engine.credentials.Credentials):
+            self.refresh_token_f = \
+                self._refresh_token_using_compute_credentials
+        else:
+            self.refresh_token_f = \
+                self._refresh_token_using_service_account_credentials
```
descent3me, line 112: I realize it may have been misleading in my previous comment when I said in "init", since it is actually a function called from init, but what I was trying to ask is if it would it be possible to set this function in `_load_credentials` to avoid the use of `isinstance`, since we already know what type of credentials we have at that point.

For example, just above the following lines:
- https://github.com/spotify/gordon-gcp/pull/118/files#diff-1f1b3f74e2435d498631282277981099L128
- https://github.com/spotify/gordon-gcp/pull/118/files#diff-1f1b3f74e2435d498631282277981099L133

tests/unit/clients/test_auth.py
```diff
@@ -252,6 +273,21 @@ async def test_refresh_token_with_app_default_cred(client_with_app_default_cred,
     assert 2 == len(caplog.records)
 
 
+@pytest.mark.asyncio
+async def test_refresh_token_with_compute_engine_cred(
+        client_with_compute_engine_cred, mock_parse_expiry,
+        caplog, payload_resp_refresh_token):
+    """Successfully refresh access token with compute engine credentials ."""
+    url = "http://metadata.google.internal/" \
+          "computeMetadata/v1/instance/service-accounts/default/token"
```
descent3me, line 282: Also quotes here, as well as parentheses for the string concatenation

tests/unit/clients/test_auth.py
```diff
@@ -55,6 +56,19 @@ def mock_oauth2_credentials(mocker, monkeypatch, fake_keyfile_data):
     return mock_creds
 
 
+@pytest.fixture
+def mock_compute_engine_credentials(mocker, monkeypatch, fake_keyfile_data):
+    mock_creds = mocker.MagicMock(gauth)
+
+    ce_creds = mocker.MagicMock(compute_engine.credentials.Credentials)
+    ce_creds._service_account_email = "default"
```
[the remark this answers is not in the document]
reneweb: Argh yep, missed those as well 😬

tests/unit/clients/test_auth.py
reneweb: yep will do

src/gordon_gcp/clients/auth.py
```diff
@@ -178,14 +185,33 @@ async def refresh_token(self):
...
+        headers = {'metadata-flavor', 'google'}
+        await self._execute_refresh_token_request(
+            lambda: self._session.get(url, headers=headers), 'GET', url)
+
+    async def _refresh_token_using_service_account_credentials(self):
         url, headers, body = self._setup_token_request()
+        await self._execute_refresh_token_request(
+            lambda: self._session.post(url, headers=headers, data=body),
+            'POST',
+            url)
```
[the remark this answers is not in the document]
reneweb: yeah, that would work and seems a bit cleaner. I'll change it

src/gordon_gcp/clients/auth.py
reneweb: Ah ok, I understand. When not having a service account file set then the default credentials code is used:
```python
        credentials, _ = gauth.default(
            scopes=['https://www.googleapis.com/auth/userinfo.email'])
```
The problem is that depending on certain conditions (gcloud available, running on compute engine or not, etc) the credentials object returned will be different. For example, when running locally it would use gcloud to return the normal service account credentials with client id, client secret, refresh token, token uri and then we want to use the token uri (normal oauth refresh token call). 
When it returns the compute engine credentials because the service runs on GKE/GCE then we want to call the metadata service. So, the logic of which credentials object we get is hidden in the `gauth.default` call and it seems the only real way to know which one it is, is checking the type.

descent3me: Aha, I didn't realize there could be two relevant environments in that section. Thanks for the detailed info! I was a little confused when this would be used since all plugins currently require keyfiles [1] [2], but I guess local gcloud auth could be handy for local debugging once we start to move them off that requirement, so it makes sense to support it.

[1] https://gordon-dns-gcp.readthedocs.io/en/latest/config-service.html#cmdoption-arg-keyfile
[2] https://gordon-dns-gcp.readthedocs.io/en/latest/config-janitor.html#cmdoption-arg-keyfile

PR #120, opened 2020-05-06
Ensure to use correct email when calling token url

reneweb:
This will make another call to the metadata API when using compute engine credentials to first ensure to get the correct email. Then this email is used to get a new access token. Previously this was not done causing the call to be made using `default` as the email, which the metadata server might reject and return a 403.

### DIFF
--- a/src/gordon_gcp/clients/auth.py
+++ b/src/gordon_gcp/clients/auth.py
@@ -198,9 +198,10 @@ class GAuthClient:
                   '?recursive=true')
 
         sa_response = await self._execute_request(sa_url, 'GET', headers)
-        email = (sa_response['email']
-                 if 'email' in sa_response
-                 else self.creds._service_account_email)
+        if 'email' in sa_response:
+            email = sa_response['email']
+        else:
+            email = self.creds._service_account_email
 
         token_url = (f'http://{metadata_host}/computeMetadata/'
                      'v1/instance/service-accounts/'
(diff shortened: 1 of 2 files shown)


src/gordon_gcp/clients/auth.py
```diff
@@ -190,17 +190,30 @@ async def refresh_token(self):
...
+        headers = {'Metadata-Flavor': 'Google'}
+        sa_url = (f'http://{metadata_host}/computeMetadata/'
+                  'v1/instance/service-accounts/'
+                  f'{self.creds._service_account_email}/'
+                  '?recursive=true')
+
+        sa_response = await self._execute_request(sa_url, 'GET', headers)
+        email = (sa_response['email']
+                 if 'email' in sa_response
+                 else self.creds._service_account_email)
```
descent3me, line 203: this might be clearer as a standard if-block. i think we lose any of the conciseness we might have gotten from a ternary operator when it ends up extending to same number of lines.

src/gordon_gcp/clients/auth.py
```diff
@@ -224,12 +237,14 @@ async def _execute_refresh_token_request(
...
-            except KeyError:
-                msg = '[{request_id}] No access token in response.'
-                logging.error(msg)
-                raise exceptions.GCPAuthError(msg)
+            return await resp.json()
+
+    def _handle_refresh_token_response(self, response):
+        try:
+            self.token = response['access_token']
+        except KeyError:
```
descent3me, line 245: I realize this wasn't changed in this PR, but would you mind switching to `if 'access_token' in response`? Key lookups are less expensive than exceptions.

src/gordon_gcp/clients/auth.py
reneweb: Yeah, the else just reasign the same as before. I think depending on the state this can work sometimes, but I mostly just took that approach by looking at the refresh function of the compute engine credentials object, where they do the same thing.

reneweb:
Fixed the issue and extend the PR comment a bit

reneweb:
Sure

PR #123, opened 2020-05-29
make keyfile optional

reneweb:
The GAuthClient can now work with default credentials, so it is not required to have the keyfile provided anymore.


ocsw:
LGTM, although I don't know if there have been any discussions.

descent3me:
@reneweb docs need to be updated as well

reneweb:
@descent3me You mean marking the keyfile in those examples as optional https://github.com/spotify/gordon-gcp/blob/43935bf6ecda74e8a333adad19beac250b167242/src/gordon_gcp/plugins/janitor/gpubsub_publisher.py#L41 right?

descent3me:
@reneweb yep!

PR #125, opened 2020-06-11
Update docs that keyfile is optional

reneweb:
Update docs to reflect previous changes (the keyfile being optional now)


reneweb:
Ah, I was just searching in the src directory, thats why I didn't see them

reneweb:
Updated those as well

PR #127, opened 2021-04-07
Update aiohttp to 3.7.4

ocsw:
Fixed version of https://github.com/spotify/gordon-gcp/pull/126

### DIFF
--- a/tests/unit/clients/test_gce.py
+++ b/tests/unit/clients/test_gce.py
@@ -118,7 +118,3 @@ class TestGCEClient:
         expected_results = compute_rsp['items']['us-west1-z']['instances'] \
             + page2['items']['us-west1-z']['instances']
         assert expected_results == results
-        requests = list(m.requests.keys())
-        assert 2 == len(requests)
-        assert ('get', filter_url) == requests[0]
-        assert ('get', url_with_token) == requests[1]
--- a/tests/unit/clients/test_gcrm.py
+++ b/tests/unit/clients/test_gcrm.py
@@ -93,8 +93,3 @@ class TestGCRMClient:
         expected_rsp = crm_one_page_rsp['projects']
         expected_rsp.append(page2['projects'].pop(0))
         assert expected_rsp == results
-        # assert requests made and their sequence
-        requests = list(m.requests.keys())
-        assert 2 == len(requests)
-        assert ('get', url_with_pagesize) == requests[0]
-        assert ('get', url_with_token) == requests[1]
--- a/tests/unit/clients/test_http.py
+++ b/tests/unit/clients/test_http.py
@@ -31,8 +31,9 @@ from tests.unit import conftest
 #####
 # Tests for simple client instantiation
 #####
+@pytest.mark.asyncio
 @pytest.mark.parametrize('provide_session', [True, False])
-def test_http_client_default(provide_session, mocker):
+async def test_http_client_default(provide_session, mocker):
     """AIOConnection is created with expected attributes."""
     session = None
     if provide_session:
@@ -51,15 +52,15 @@ def test_http_client_default(provide_session, mocker):
         assert auth_client._session is client._session
         assert session is not client._session
 
-    client._session.close()
+    await auth_client._session.close()
 
 
 @pytest.fixture
-def client(mocker, auth_client):
+async def client(mocker, auth_client):
     session = aiohttp.ClientSession()
     client = http.AIOConnection(auth_client=auth_client, session=session)
     yield client
-    session.close()
+    await session.close()
 
 
 args = 'token,expiry,is_valid_token_expected'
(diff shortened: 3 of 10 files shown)


tests/unit/clients/test_gce.py
```diff
@@ -120,5 +121,5 @@ async def test_list_instances_retrieves_multiple_pages(
         assert expected_results == results
         requests = list(m.requests.keys())
         assert 2 == len(requests)
-        assert ('get', filter_url) == requests[0]
-        assert ('get', url_with_token) == requests[1]
+        assert ('get', yarl.URL(filter_url)) == requests[0]
+        assert ('get', yarl.URL(url_with_token)) == requests[1]
```
descent3me, line 125: It feels like we're testing the wrong things here. We don't really care about the data type, only that the correct urls were called, which we already kind of know, since we only mocked the two endpoints and the results wouldn't match if the mock response data wasn't used.

I would be fine dropping these asserts altogether, since they don't seem to give anything valuable, and seem more to be testing the internals of aioresponses. If you absolutely want to keep them, I would change them to compare strings, because the yarl URL type is a side effect of the mocking library, not what we expect about the GCE client.

tests/unit/plugins/janitor/test_reconciler.py
```diff
@@ -180,13 +180,8 @@ async def validate_rrsets_by_zone():
     await recon_client.cleanup()
 
     assert exp_log_records == len(caplog.records)
-    if exp_log_records == 2:
-        # it's in a cancelling state which can't be directly tested
-        assert not coro1.done()
-        assert not coro2.done()
-    else:
-        assert coro1.done()
-        assert coro2.done()
```
descent3me, line 189: I don't totally understand what this test was trying to test either, but if you do, +1

tests/unit/clients/test_gce.py
ocsw: I didn't look at what the tests are doing, just getting them to pass.  This isn't testing the yarl.URL()s, those tuples are actually the keys in the requests dict.  It's true, though, that testing these isn't super valid...

tests/unit/plugins/janitor/test_reconciler.py
ocsw: No, not really.  Just the behavior apparently changed wrt something about cancelling requests.

tests/unit/clients/test_gce.py
descent3me: What I mean by testing the `yarl.URL` type is that, because the syntax is `assert expected == actual`, we're saying we expect and are testing that this is object is of the type `yarl.URL`, as opposed to something like `assert filter_url == str(requests[0][1])` which allows our expected value to more closely match what we actually expect.

However I don't necessarily want to suggest that as my other point was that it seems like `requests` is actually an internal detail of `aioresponses` since it changed in a minor version, doesn't appear in the public docs, the data structure is awkward to use, and the way we're using it (casting a dictionary keyset to list to test order in < Python 3.7) is a bit strange as well.

ocsw:
@descent3me Ok, how about now?

tests/unit/clients/test_gce.py
```diff
@@ -118,7 +118,3 @@ async def test_list_instances_retrieves_multiple_pages(
         expected_results = compute_rsp['items']['us-west1-z']['instances'] \
             + page2['items']['us-west1-z']['instances']
         assert expected_results == results
-        requests = list(m.requests.keys())
```
ocsw, line 121: This isn't really testing anything useful

tests/unit/clients/test_gcrm.py
```diff
@@ -93,8 +93,3 @@ async def test_list_all_active_projects_multiple_pages(
         expected_rsp = crm_one_page_rsp['projects']
         expected_rsp.append(page2['projects'].pop(0))
         assert expected_rsp == results
-        # assert requests made and their sequence
```
ocsw, line 96: This isn't really testing anything useful

tests/unit/clients/test_http.py
```diff
@@ -119,7 +121,7 @@ async def mock_valid_token_set():
     assert resp == resp_text
 
     assert 1 == mock_refresh_token_called
-    request = mocked.requests[('get', conftest.API_URL)][0]
+    request = mocked.requests[('get', yarl.URL(conftest.API_URL))][0]
```
descent3me, line 124: This is the only bit left. If we have to keep something like it (since I'm guessing you don't want to refactor http client to make testing default header addition easier), can we do something like `request = mocked.requests.popitem()[1][0]` instead? I know it's not super pretty, but I think it should be the only item, and it seems weird to do requiring a whole library for one already not-so-pretty line.

ocsw: I'm not sure I get your suggestion.  requests is a dict.  so surely it's popitem()[0]?  but also that's very much making an assumption that there's exactly one item in the dict

ocsw: I guess I could test len first

ocsw: nm, I didn't know how popitem() works, apparently

ocsw: how's this?

descent3me (approved):
Thanks!

